PR #51, opened 2015-02-26
refactor build.sh

oilbeater:
The way you parse arguments by echo and grep really amused me :)
Though I am not proficient in shell, however by using optget it looks better than the previous.
What's more, the statement "echo 'docker push rancher/ubuntuconsole'"  seems in a wrong place.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -9,13 +9,25 @@ export DOCKER_IMAGE=rancher-os-build
 mkdir -p dist
 docker run --rm -it -e CHOWN_ID=$(id -u) -v $(pwd)/dist:/source/target $DOCKER_IMAGE
 
-# Stupidest argparse ever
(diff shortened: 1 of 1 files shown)


ibuildthecloud:
Thanks for the proper arg parsing.  Sorry it took so long to review this.  My only comment is that the order the of program args changes the execution of the script.  For example:

   ./build.sh --images --push

Is different from

    ./build.sh --push --images

Those two command should do the same thing, but right now the order the args determines the order of the operations in the script.

oilbeater:
To find a elegant and simple way to solve the problem seems difficult by shell @ibuildthecloud

ibuildthecloud:
@oilbeater I'm going to close this PR.  If you find a better approach please reopen or create another PR. Thanks

PR #52, opened 2015-02-26
notify if eth0 not found

sidharthamani:
@ibuildthecloud

### DIFF
--- a/scripts/dockerimages/scripts/network.sh
+++ b/scripts/dockerimages/scripts/network.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+if [ ! -d "/sys/class/net/eth0" ]; then
+    echo "eth0 not found, ssh access cannot be provided!"
+fi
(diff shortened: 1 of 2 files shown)


scripts/dockerimages/scripts/network.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+if [ ! -d "/sys/class/net/eth0" ]; then
+    echo "eth0 not found, ssh access cannot be provided!"
```
cloudnautique: The problem is larger then ssh. Should probably say network unavailable... should the script just exit? Since we won't get a dhcp address.

ibuildthecloud:
I don't think this actually provides much value as that message won't be echo'd on the screen

ibuildthecloud:
I'd really like to know why udev doesn't detect the device.  Have you found out why that is?

PR #67, opened 2015-03-03
ubuntu-console: use latest Ubuntu 14.04 image

imikushin:
see https://registry.hub.docker.com/_/ubuntu/


ibuildthecloud:
Thanks

ibuildthecloud:
Oh sorry, I read the commit backward.  The reason that we use 14.04.1 as opposed to 14.04 is for reproducibility.  We essentially pin all versions and don't want to use labels that are known to change (like latest).  Having said that, 14.04.2 is the latest and I'll gladly accept a PR that updates it to 14.04.2

imikushin:
pinned 14.04.2 in the next commit

PR #76, opened 2015-03-06
etcd and flannel addons

imikushin:
etcd and flannel addons

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@
 .dockerfile
 *.swp
 Dockerfile
+/.idea
+/*.iml
+/scripts/dockerimages/tmp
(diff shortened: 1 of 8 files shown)


ibuildthecloud:
This is really cool.  Thanks.  I will review this as soon as I get the chance.  We for sure would like to support etcd, fannel, and any other popular Docker service as an addon.

imikushin:
I just hope this runs fine.

ibuildthecloud:
@imikushin Should we close this?  As a result of the k8s containers you are working on will we also get a generic etcd and flannel one?

imikushin:
Generic etcd is possible right now (with `quay.io/coreos/etcd:v2.0.10`). Flannel (`quay.io/coreos/flannel:0.4.0`) needs variable substitution in compose to get user configuration or we'll need to repackage it to add wrapper scripts, which is easy to do. So, closing it now.

PR #91, opened 2015-03-20
Set defaults based off of Rancher OS info

cloudnautique:
If running on RancherOS gather sane defaults for users when running rancheros-install command.


scripts/dockerimages/scripts/rancheros-install
```diff
@@ -45,8 +52,8 @@ if [ "$(whoami)" != "root" ]; then
 fi
 
 if [[ -z "${INSTALL_TYPE}" ]]; then
-    echo "$0: No install type specified, -t required to install" 1>&2
-    exit 1
+    echo "No install type specified...defaulting to generic"
```
ibuildthecloud: I like the dramatic pause :)

PR #104, opened 2015-03-25
add ipv4ll auto conf

wlan0:
@ibuildthecloud

### DIFF
--- a/cmd/network/ipv4ll.go
+++ b/cmd/network/ipv4ll.go
@@ -1,0 +1,75 @@
+package network
+
+import (
+	"encoding/binary"
+	"fmt"
+	"math/rand"
+	"net"
+
+	log "github.com/Sirupsen/logrus"
+
+	"github.com/j-keck/arping"
+	"github.com/vishvananda/netlink"
+)
+
+func AssignLinkLocalIP(link netlink.Link) error {
+	ifaceName := link.Attrs().Name
+	iface, err := net.InterfaceByName(ifaceName)
+	if err != nil {
+		log.Error("could not get information about interface")
+		return err
+	}
+	addrs, err := iface.Addrs()
+	if err != nil {
+		log.Error("Error fetching existing ip on interface")
+	}
+	for _, addr := range addrs {
+		if addr.String()[:7] == "169.254" {
+			log.Info("Link Local IP already set on interface")
+			return nil
+		}
+	}
+	randSource, err := getPseudoRandomGenerator(link.Attrs().HardwareAddr)
+	if err != nil {
+		return err
+	}
+	// try a random address upto 10 times
+	for i := 0; i < 10; i++ {
+		randGenerator := rand.New(*randSource)
+		randomNum := randGenerator.Uint32()
+		dstIP := getNewIPV4LLAddr(randomNum)
+		if dstIP[2] == 0 || dstIP[2] == 255 {
+			i--
+			continue
+		}
+		_, _, err := arping.PingOverIfaceByName(dstIP, ifaceName)
+		if err != nil {
+			// this ip is not being used
+			addr, err := netlink.ParseAddr(dstIP.String() + "/16")
+			if err != nil {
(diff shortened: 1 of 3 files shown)


cmd/network/ipv4ll.go
```diff
@@ -0,0 +1,79 @@
...
+			log.Infof("Set %s on %s", dstIP.String(), link.Attrs().Name)
+			return nil
+		}
+	}
+	log.Error("Could not find a suitable ipv4ll")
+	return fmt.Errorf("Could not find a suitable ipv4ll")
+}
+
+func getNewIPV4LLAddr(randomNum uint32) net.IP {
+	byte1 := randomNum & 255 // use least significant 8 bits
```
ibuildthecloud: You need to make sure you never use 169.254.255.255 or 169.254.0.0

vincent99: All of 169.254.0.* and 169.254.255.* are not valid:  http://tools.ietf.org/html/rfc3927#section-2.1

cmd/network/ipv4ll.go
```diff
@@ -0,0 +1,79 @@
...
+	byte2 := randomNum >> 24 // use most significant 8 bits
+	return []byte{169, 254, byte(byte1), byte(byte2)}
+}
+
+func getPseudoRandomGenerator(ifaceName string) (*rand.Source, error) {
+	haAddr, err := getHaAddr(ifaceName)
```
ibuildthecloud: You don't have to jump through so many hoops to do this.  The hardware address is already available to you from the `link.Attr().HardwareAddr`.  The hardware address is really just a `[]byte` so you can just convert that to a in64 with binary.Varint and use that as the seed.

cmd/network/ipv4ll.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		log.Error("Could not obtain MAC address")
+		src := rand.NewSource(0)
+		return &src, err
+	}
+	seed := generateSeed(haAddr)
+	fmt.Println(seed)
```
ibuildthecloud: don't println

cmd/network/ipv4ll.go
```diff
@@ -0,0 +1,79 @@
...
+		dstIP := getNewIPV4LLAddr(randomNum)
+		_, _, err := arping.PingOverIfaceByName(dstIP, ifaceName)
+		if err != nil {
+			// this ip is not being used
+			addr, err := netlink.ParseAddr(dstIP.String() + "/16")
+			if err != nil {
+				log.Errorf("error while parsing ipv4ll addr, err = %v", err)
+				return err
+			}
+			if err := netlink.AddrAdd(link, addr); err != nil {
```
ibuildthecloud: What happens if I run this over and over again.  Do I end up assigning a random LL IP each time?  You should probably check to see if a 169.254/16 is already assigned.

wlan0: You get the same IP if you run the program again... If you run the for...loop again though, you get a new IP (each time I do `rand.New(*randSource)`)

cmd/network/ipv4ll.go
wlan0: thanks @vincent99

cmd/network/ipv4ll.go
ibuildthecloud: If you already have the IP assigned to the NIC the arping will still fail?

wlan0: Yes, It fails

cmd/network/ipv4ll.go
wlan0: When I wrote it initially, I intended it be generic. Once I integrated it with netconf, realized I dont need it to be. I'll change this

wlan0:
@ibuildthecloud updated the code

wlan0:
@ibuildthecloud updated the code

ibuildthecloud:
LGTM.  I'll manually merge this so we can get the godep updates too

PR #105, opened 2015-03-26
When in Debug mode show more logs

cloudnautique:
When in debug=true dump the log output of the container to
stdout/stderror so that it can be read from console on boot. This
is particularly helpful when troubleshooting installation problems
over a serial console.

### DIFF
--- a/init/sysinit.go
+++ b/init/sysinit.go
@@ -132,6 +132,10 @@ func runContainersFrom(startFrom string, cfg *config.Config, containerConfigs []
 				log.Errorf("Failed to run %v: %v", containerConfig.Id, container.Err)
 			}
 
+			if cfg.Debug {
+				container.Log(os.Stdout, os.Stderr, false)
+			}
+
 			if containerConfig.ReloadConfig {
 				log.Info("Reloading configuration")
 				err := cfg.Reload()
@@ -158,11 +162,6 @@ func tailConsole(cfg *config.Config) error {
 		return nil
 	}
 
-	client, err := docker.NewSystemClient()
-	if err != nil {
-		return err
-	}
-
 	for _, container := range cfg.SystemContainers {
 		if container.Id != config.CONSOLE_CONTAINER {
 			continue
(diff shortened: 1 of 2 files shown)


ibuildthecloud:
When one creates a new docker.Container you pass in the DOCKER_HOST string.  This way that container can be a "system" container or a "user" container.  I'd like to make this a bit more generic so it can be useful for the user container support I'm adding.

Can you make a `func (*c Container) Log(stdout io.Writer, stderr io.Writer) error` so then in `sys-init` you can just do `go container.Log(os.Stdout, os.Stderr)`.

Oh yeah, also Logs() is a blocking call so you want to put it in a goroutine.

init/sysinit.go
```diff
@@ -132,6 +132,10 @@ func runContainersFrom(startFrom string, cfg *config.Config, containerConfigs []
 				log.Errorf("Failed to run %v: %v", containerConfig.Id, container.Err)
 			}
 
+			if cfg.Debug {
+				container.Log(os.Stdout, os.Stderr)
```
ibuildthecloud: I understand that you want blocking behavior to see the output of the containers sequentially, but the problem ends up being that not all containers stop.  For example userdocker container doesn't stop.  I believe what will happen is that it will just stop and wait for the user docker container to finish and the boot will hang.

ibuildthecloud: An alternative to a global debug flag could be that we have an option of which service we want to log to the console.  For example `rancher.log.containers=[network]`

cloudnautique: As implemented right now, its not following. It just dumps the logs and exits, similar to the docker cli when there is a long running process.

For readability in the startup process, I'd like to make the call blocking. So I'd like to take out the go routine in the container.Log method. If we are going to keep the method as part of the Container type, I think we should add a boolean to follow, but not enable it here. 

Thoughts?

ibuildthecloud: Sounds good, remove the go routine and add a follow option to Log

ibuildthecloud:
After the refactor to use compose syntax the debug logic should go it `docker/factory.go`

ibuildthecloud:
Is this still valid?  Can it be updated to work with v0.3.0?

cloudnautique:
I think so. Need to determine what happens when we turn on syslog and debug. If syslog is enabled we stop outputting to the console, so we need to change that. Unless I'm missing something.

ibuildthecloud:
In 0.4 now you can add `rancher.log` to the `/proc/cmdline` and all services will be logged to the console.  I believe this addresses this use case.

PR #122, opened 2015-03-31
use new os-base with dhcpcd

wlan0:
@ibuildthecloud

### DIFF
--- a/config/default.go
+++ b/config/default.go
@@ -164,7 +164,7 @@ func NewConfig() *Config {
 				Id: "network",
 				Cmd: "--name=network " +
 					"--rm " +
-					"--cap-add=NET_ADMIN " +
+					"--privileged " +
 					"--net=host " +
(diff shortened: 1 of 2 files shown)


config/default.go
```diff
@@ -164,7 +164,7 @@ func NewConfig() *Config {
 				Id: "network",
 				Cmd: "--name=network " +
 					"--rm " +
-					"--cap-add=NET_ADMIN " +
+					"--privileged " +
```
wlan0: NET_ADMIN is not enough for dhcpcd to work.

strace'ing dhcpcd shows that it tries to write into proc fs

```
open("/proc/sys/net/ipv4/conf/eth0/promote_secondaries", O_WRONLY|O_CREAT|O_TRUNC, 0666) = -1 EROFS (Read-only file system)
```

So, i added the privileged flag

ibuildthecloud:
Manually merged to master

PR #125, opened 2015-04-02
Add OEM support to network dhcp container.

cloudnautique:
In order to get the networking configuration to work on GCE, we
needed to add a custom network interface configuration. This change
makes it possible to do so, and if we need other providers.

This is likely deprecated by PR #122


cloudnautique:
This is not needed because of #122

PR #145, opened 2015-04-10
bump IMAGE_VERSION

imikushin:
fix: rancher/ubuntu-console is still at v0.2.1 and is 
inconsistent in /etc/passwd with the base


ibuildthecloud:
Thanks.  We're about to do v0.3.0-rc2 so I bumped master to rc2 instead.  The ubuntuconsole:v0.3.0-rc2 image is pushed to docker hub so you can run off of master if you choose.

PR #158, opened 2015-04-11
Add 'rancherctl os version' command

darnould:
Adds `rancherctl os version` command to show the currently installed version of RancherOS.


darnould:
Addresses the same core issue raised separately at https://github.com/rancherio/os/issues/148.

ibuildthecloud:
I like this.  For most scripting things we've been using "rancherctl -v" but that requires you to parse the last argument.  This is much cleaner.

PR #163, opened 2015-04-13
Update os.go

deniseschannon:
Moved the printout of "Upgrading to rancher/os:v.xxx" so that it only prints out if it's not staging.

Fixes https://github.com/rancherio/os/issues/152

### DIFF
--- a/cmd/control/os.go
+++ b/cmd/control/os.go
@@ -165,9 +165,10 @@ func startUpgradeContainer(image string, stage, force bool) {
 		log.Fatal(container.Err)
 	}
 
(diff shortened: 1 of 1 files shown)


wlan0:
:+1:  LGTM

cmd/control/os.go
```diff
@@ -165,9 +165,10 @@ func startUpgradeContainer(image string, stage, force bool) {
 		log.Fatal(container.Err)
 	}
 
-	fmt.Printf("Upgrading to %s\n", image)
-
+
```
wlan0: This is not very important, but we should remove the new line here

Issue #173, opened 2015-04-14
Instance hangs on boot


cloudnautique:
During boot the instance hangs on the console logs

[11/14] [userdockerwait]: Started

The behavior has been observed in AWS and GCE with instances created from ibuildthecloud/WIP.

sheng-liang:
The real problem is we don't print what's running when boot hangs.

deniseschannon:
Closing due to all the changes in RancherOS.

Issue #171, opened 2015-04-14
Network device name during bootup


linusyong:
The issue as describe below:
* ISO file: rancheros.iso 0.3.0-rc2
* Environment: VirtualBox 4.3.12 r93733 Windows
* Network: bridge or NAT (both tested)

During first time boot up, `ifconfig` does not show any ethernet device.  `/var/log/messages` shows that ethernet device has been renamed using udev predictable name:
```
Apr  13 05:07:21 rancher kernel: [    5.633363] e1000 0000:00:03.0 enp0s3: renamed from eth0
```

`ifconfig enp0s3` shows the device is there and `dhcpcd enp0s3` gives the device IP.

deniseschannon:
This is only related to v0.3.0-rc2:

For a temporary fix:

In the VM in VirtualBox, go to Network Settings -> Advanced Settings.

Change "Adapter Type" to "Paravirtualized Network (virtio-net)".

deniseschannon:
@linusyong This is fixed in v0.3.0-rc2. Please try with that version and see if this issue is happening.

deniseschannon:
Fixed in v0.3.0

PR #174, opened 2015-04-14
Disable predictable naming.

cloudnautique:
When we upgraded udev, we also started using predictable naming for
interfaces. This has caused connectivity issues in a variety of
environments. Disabling for now.

For background:
http://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/

Issues stemming from this:
#171 
#173 (will address in OS installer)

### DIFF
--- a/scripts/package
+++ b/scripts/package
@@ -30,7 +30,7 @@ default rancheros
 label rancheros
     kernel /boot/vmlinuz
     initrd /boot/initrd
-    append quiet rancher.password=rancher
+    append quiet rancher.password=rancher net.ifnames=0
 EOF
 
 cd ${CD}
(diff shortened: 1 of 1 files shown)


wlan0:
@cloudnautique Wouldn't it be better to start using predictable naming and configure rancherOS bootup to work with the new naming scheme? The change to rancherOS is really small for that.

cloudnautique:
For now disabling in the ISO puts us into a known state. A user who wants this behavior can reenable it on installation, and create the networking config, so it is not going away.

ibuildthecloud:
Please instead revert https://github.com/rancherio/os/commit/402cbbdf6f3e6b4eb547d2d1cac3eac5bee38501

ibuildthecloud:
I'd rather globally disable it.  If people complain we can address it then.  I can't find a general consensus on which approach is better.  eth* addresses short term issues

cloudnautique:
Closing in favor of #176

PR #183, opened 2015-04-16
Use service not addon

hibooboo2:


ibuildthecloud:
No longer valid after docs refactor

PR #197, opened 2015-04-17
Needed to add a newline.

cloudnautique:
Prevents the last and first SSH Keys from Instance and projects from being added to the same line.


cloudnautique:
ping @ibuildthecloud

PR #239, opened 2015-04-27
Machine RancherOS Support

wlan0:
@ibuildthecloud  @sheng-liang

### DIFF
--- a/config/default.go
+++ b/config/default.go
@@ -323,6 +323,21 @@ func NewConfig() *Config {
 				Ipc:     "host",
 				Net:     "host",
 			},
+			"acpid": {
+				Image:      "acpid",
+				Privileged: true,
+				Links: []string{
+					"console",
+				},
+				Labels: []string{
+					SCOPE + "=" + SYSTEM,
+				},
+				VolumesFrom: []string{
+					"command-volumes",
+					"system-volumes",
+				},
+				Net: "host",
+			},
 		},
 		ServicesInclude: map[string]bool{
 			"ubuntu-console": false,
--- a/init/bootstrap.go
+++ b/init/bootstrap.go
@@ -3,6 +3,7 @@ package init
 import (
 	"os"
 	"os/exec"
+	"strings"
 	"syscall"
 
 	log "github.com/Sirupsen/logrus"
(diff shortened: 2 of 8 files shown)


wlan0:
@ibuildthecloud  @sheng-liang -- ACPID daemon does not receive events inside a container

The issue has something to do with the mount namespaces, you can see it running though. 

I've updated the code.

wlan0:
I propose that we do not containerize acpid for now. It is really not required to get rancheros ready for machine. 

We can revisit this, once we have a working machine build.

@ibuildthecloud  @sheng-liang

wlan0:
Also note, this PR requires OS base with acpid support

config/default.go
```diff
@@ -60,7 +60,7 @@ func NewConfig() *Config {
 				Nameservers: []string{"8.8.8.8", "8.8.4.4"},
 			},
 			Interfaces: map[string]InterfaceConfig{
-				"eth0": {
+				"eth*": {
```
ibuildthecloud: Why are you doing changing this?

init/bootstrap.go
```diff
@@ -43,10 +46,18 @@ outer:
 			continue
 		}
 
-		for _, b := range buffer {
-			if b != 0 {
-				log.Infof("%s not empty", dev)
-				continue outer
+		boot2docker := false
+
+		if strings.HasPrefix(string(buffer[:30]), boot2dockerMagic) {
```
ibuildthecloud: please do len(b2dmagic) instead of 30

config/default.go
wlan0: machine for virtualbox configures two interfaces.

scripts/dockerimages/scripts/console.sh
```diff
@@ -33,6 +33,15 @@ if [ ! -d ${RANCHER_HOME} ]; then
...
 fi
 
+DOCKER_HOME=/home/docker
+if [ ! -d ${DOCKER_HOME} ]; then
+    mkdir -p ${DOCKER_HOME}
+    chown docker:docker ${DOCKER_HOME}
+    chmod 2755 ${DOCKER_HOME}
+fi
+
+/usr/sbin/acpid
```
ibuildthecloud: This needs to be a separate container?

wlan0: No, it need not be.

PR #251, opened 2015-04-29
make modules settable through rancherctl config

wlan0:
Fixes: https://github.com/rancherio/os/issues/235

### DIFF
--- a/init/init.go
+++ b/init/init.go
@@ -357,6 +357,7 @@ func RunInit() error {
 		func(cfg *config.Config) error {
 			return cfg.Reload()
 		},
+		loadModules,
 		setResolvConf,
 		func(cfg *config.Config) error {
 			return createDirs(postDirs...)


wlan0:
@sheng-liang @ibuildthecloud

init/init.go
```diff
@@ -349,7 +349,6 @@ func RunInit() error {
 		},
 		createGroups,
 		extractModules,
-		loadModules,
```
ibuildthecloud: We can't remove this here.  We need this incase you need a module to mount the disk.  We can just have it twice.  Just make sure if you load a module twice it doesn't fail.

config/default.go
```diff
@@ -38,7 +38,9 @@ func NewConfig() *Config {
 				"-H", DOCKER_SYSTEM_HOST,
 			},
 		},
-		Modules: []string{},
+		Modules: []string{
```
ibuildthecloud: I know @sheng-liang asked for this, but I disagree.  I think we should just compile tun into the default kernel.  I'm pretty sure if you were to swap out a kernel that didn't have tun, then our default configuration will fail because it can't load tun.

wlan0:
@ibuildthecloud @sheng-liang Updated the code to load modules twice.

PR #271, opened 2015-05-01
Add doc for service delete

wlan0:
@deniseschannon @ibuildthecloud

### DIFF
--- a/docs/rancherctl/service/index.md
+++ b/docs/rancherctl/service/index.md
@@ -18,6 +18,7 @@ _Available as of v0.3.0._
 |`enable`	| Turn on an service|
 |`disable`	|Turn off an service|
 |`list`	|	List services and state|
+|`delete`| Delete services|
 
 ### List
 ---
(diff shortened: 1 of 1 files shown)


docs/rancherctl/service/index.md
```diff
@@ -59,3 +60,17 @@ $ sudo rancherctl service list
...
+```
+$ sudo rancherctl service add etcd.yml
+$ sudo rancherctl service list
+enabled  etcd.yml
+disabled ubuntu-console
+$ sudo rancherctl delete etcd.yml
+$ sudo rancherctl service list
+disabled ubuntu-console
+```
+Note that you **cannot** delete the ubuntu-console , because that was compiled into RancherOS
```
ibuildthecloud: Maybe we should not compile that in anymore.  I don't see any value of doing that.

docs/rancherctl/service/index.md
```diff
@@ -59,3 +60,17 @@ $ sudo rancherctl service list
 disabled ubuntu-console
 ```
 
+### Delete
+---
+The `delete` command can be used to delete services that were added dynamically. i.e. All the services that were not compiled in can be deleted. 
+
+```
+$ sudo rancherctl service add etcd.yml
```
ibuildthecloud: I think the command is enable, not add

docs/rancherctl/service/index.md
wlan0: Thats a topic for discussion.

deniseschannon: Instead of saying what's compiled into RancherOS? Wouldn't it be more about what's in os-services repo? Can you remove a service that's in the os-services repository?

ibuildthecloud: Yes, you can remove what is in os-services repo.  We don't actually remove the service, just the reference too it in rancher.yml

ibuildthecloud: @wlan0 Can you try removing ubuntu-console from config/default.go and see if there are any side effect?  It seems like there is no reason for it to be there now that we have os-service repo.

deniseschannon:
Added this content with the new website structure PR #284.

PR #277, opened 2015-05-01
fix service restart

### LINKED ISSUE #230: `restart: always` not applied to services in compose format
Repro:
  1. `rancherctl s enable https://gist.githubusercontent.com/imikushin/fec33701294b77641022/raw/d6c488f0fb1cb21d60e9d8b64c5a1e4abff767fe/nginx.yml`
  2. Reboot.
  3. Wait until `nginx` container is started. `docker ps -a` shows the container is running.
  4. Simulate container main process failure:  
`ps -ef | grep 'nginx: master' | awk '{print $2}' | xargs -I{} kill {}`
  5. `docker ps -a` shows `nginx` container is NOT running.

If I just use docker CLI to run a container with `--restart=always` and kill its main process from outside, the container gets restarted as expected.

imikushin:
fixes #230

### DIFF
--- a/Godeps/_workspace/src/github.com/rancherio/rancher-compose/docker/factory.go
+++ b/Godeps/_workspace/src/github.com/rancherio/rancher-compose/docker/factory.go
@@ -11,10 +11,6 @@ import (
 )
 
 func Convert(c *project.ServiceConfig) (*runconfig.Config, *runconfig.HostConfig, error) {
-	volumes := map[string]struct{}{}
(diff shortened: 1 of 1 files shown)


ibuildthecloud:
@imikushin I have updated all of the Godeps, including the latest rancher-compose.  For now if you need to change rancher-compose code put in a PR to rancherio/rancher-compose repo in the refactor branch.  By next wednesday it will be master and a bit less confusing.

PR #280, opened 2015-05-01
Add better power function support

wlan0:
@ibuildthecloud @sheng-liang 
Fixes https://github.com/rancherio/os/issues/178


cmd/power/power.go
```diff
@@ -89,19 +138,7 @@ func runDocker(name string) error {
 		})
 	}()
 
-	err = client.StartContainer(powerContainer.ID, powerContainer.HostConfig)
-	if err != nil {
-		return err
-	}
-
-	_, err = client.WaitContainer(powerContainer.ID)
```
ibuildthecloud: Can you explain why the wait was removed?

cmd/power/power.go
```diff
@@ -114,23 +151,41 @@ func common(name string) {
...
-func PowerOff() {
+func PowerOff(shutdown bool) {
 	common("poweroff")
-	reboot(syscall.LINUX_REBOOT_CMD_POWER_OFF)
+	if os.ExpandEnv("${IN_DOCKER}") == "true" {
+		PowerOff2(0, shutdown)
+	}
+}
+
+func PowerOff2(timeout int, shutdown bool) {
```
ibuildthecloud: We are not maintaining any API compatibility, we don't need to create "2" functions

cmd/power/power.go
```diff
@@ -19,6 +21,53 @@ const (
...
+	app.EnableBashCompletion = true
+	app.Action = power
+	app.HideHelp = true
+	app.Flags = []cli.Flag{
+		cli.BoolFlag{
+			Name:  "f",
+			Usage: "Does not invoke shutdown, and instead performs the actual action",
+		},
+		cli.BoolFlag{
+			Name:  "n",
```
ibuildthecloud: Does -n actually do anything in this code?

cmd/power/power.go
```diff
@@ -19,6 +21,53 @@ const (
 	DOCKER_CGROUPS_FILE = "/proc/self/cgroup"
 )
 
+func PowerMain() {
+	app := cli.NewApp()
+
+	app.Name = os.Args[0]
+	app.Version = config.VERSION
+	app.Author = "Rancher Labs, Inc."
+	app.Email = "sid@rancher.com"
```
ibuildthecloud: Please don't put your personal email.  Git commits/blame are better to determine authors

cmd/power/shutdown.go
```diff
@@ -17,28 +19,55 @@ func Main() {
...
+		reboot := c.Bool("r")
+		halt := c.Bool("H")
+		_ = c.Bool("P")
+		time := c.Args().First()
 
-	if reboot == "now" {
-		Reboot()
-	} else if poweroff == "now" {
-		PowerOff()
+		timeout, err := strconv.Atoi(time)
```
ibuildthecloud: timeout for shutdown is in the format "+m" or "HH:MM"

cmd/power/power.go
```diff
@@ -19,6 +21,53 @@ const (
...
+			Usage: "Brings down IDE drives before shutdown",
+		},
+	}
+	app.Run(os.Args)
+}
+
+func power(c *cli.Context) {
+	force := c.Bool("f")
+	switch os.Args[0] {
+	case "poweroff":
```
ibuildthecloud: Args[0] may equal the full path of the file also, not merely the executable name.  For example "/sbin/poweroff"

cmd/power/power.go
ibuildthecloud: I'm taking a guess at why you did this, but I think it may help to point out that `shutdown -H <time>` runs in the foreground.  It will block until the shutdown time.  If you hit `ctrl+c` shutdown will be cancelled.

wlan0: You're right. I wanted it to do background shutdown.

cmd/power/power.go
wlan0: I Just copied this over from shutdown.go. I'll remove it

cmd/power/power.go
ibuildthecloud: Okay, well shutdown doesn't run in the background.  Can you fix this

cmd/power/power.go
wlan0: No, it doesn't. The linux kernel always calls sync before shutting down

ibuildthecloud: @wlan0 The kernel doesn't call sync AFAIK.  That is why we added the Sync() call to begin with.  Passing `-n` should not run the `Sync()`.  Am I misunderstanding what `-n` does?

wlan0: The docs here talk about linux always calling sync http://manpages.ubuntu.com/manpages/gutsy/man8/halt.8.html

ibuildthecloud: `-n` means that the power function should not call `sync()` itself.  Whether the Linux kernel does sync internally is irrelevant.  If `-n` is passed we need to not call `sync()`.

cmd/power/shutdown.go
```diff
@@ -14,31 +19,84 @@ func Main() {
...
+		timeToParse = strconv.Itoa(mins)
+	} else if timeValue == "now" {
+		timeToParse = "0"
+	} else {
+		log.Fatalf("invalid time value %s", timeValue)
+	}
 	common("")
-	reboot := c.String("r")
-	poweroff := c.String("h")
+	if os.ExpandEnv("${IN_DOCKER}") == "true" {
```
ibuildthecloud: Can you test if `ctrl-c` will cancel shutdown?  It should be if you run `shutdown -h +1` and then hit `ctrl-c` it won't shutdown.  I think you may need to sleep before `common("")`.

cmd/power/power.go
```diff
@@ -197,7 +230,7 @@ func shutDownContainers() error {
 		}
 
 		log.Infof("Stopping %s : %v", container.ID[:12], container.Names)
-		stopErr := client.StopContainer(container.ID, uint(timeout))
+		stopErr := client.StopContainer(container.ID, 0)
```
ibuildthecloud: Don't put `0` here.  That means docker will just `SIGKILL` all the containers.  We need some timeout so that containers have an option to gracefully shutdown.  I think 2 or 5 seconds should be good.

wlan0:
@ibuildthecloud  Changed according to your comments

wlan0:
@ibuildthecloud  Took care of crossing through midnight

PR #282, opened 2015-05-04
fix #50

### LINKED ISSUE #50: how do I create a second network interface ?
I did not look deep into it, but it seems to be missing some network drivers/device.

imikushin:
fix #50


ibuildthecloud:
@imikushin We actually have done this on purpose.  We don't want to assume anything about the NIC so we've choose to only default to DHCP for eth0.  If somebody wants DHCP on other nics they just need to configure it in their cloud-config.

PR #290, opened 2015-05-06
stage all service on enable

wlan0:
@ibuildthecloud  This requires my changes to rancher-compose. I'll leave the changes to Godeps here as a representative of the changes I made to rancher-compose


wlan0:
Holding off on the fix for this. Will reopen when relevant

PR #297, opened 2015-05-07
Fixes creating Hash

### LINKED ISSUE #292: io.rancher.os.remove=false is broken with new labels change
With v0.3.1, 

If the label `io.rancher.os.remove=false` is applied, it doesn't work. 

`rancherserver` and `rancherstate` from previous boot become `rancherserver-f6f2b405cbb3ddcaebe1d3d21dc4d28c12182276` and `rancher-state-509585100fb8965b8e0300c47da08bc62e87f2e6` on the next boot and don't go anywhere, just stay exited.

deniseschannon:
@ibuildthecloud This should fix the hash that would fix #292.

### DIFF
--- a/docker/container.go
+++ b/docker/container.go
@@ -73,28 +73,42 @@ func getHash(containerCfg *config.ContainerConfig) (string, error) {
 		//Sort serviceKeys alphabetically
 		sort.Strings(serviceKeys)
 
-		//New slice to sort through Labels Map
-		var labelKeys []string
+		//New slice to sort through Keys
+		//var sliceKeys []string
 
 		//Go through keys and write hash
 		for i := 0; i < len(serviceKeys); i++ {
 			serviceValue := unsortedKeyValue[serviceKeys[i]]
 			serviceType := unsortedKeyType[serviceKeys[i]]
+			sliceKeys := []string{}
 
 			//Only need to check for Labels key as it's a Map and write hash after sorting the label names
 			if serviceType == reflect.TypeOf(project.NewSliceorMap(make(map[string]string))) {
 				for lkey := range serviceValue.(*project.SliceorMap).MapParts() {
-					labelKeys = append(labelKeys, lkey)
+					sliceKeys = append(sliceKeys, lkey)
 				}
-				sort.Strings(labelKeys)
+				sort.Strings(sliceKeys)
 
-				for j := 0; j < len(labelKeys); j++ {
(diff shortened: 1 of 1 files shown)


deniseschannon:
@ibuildthecloud The hashing has been corrected to incorporate changes for maps and slices (when content of slice is the same, but in a different order). There is something strange going on with starting/stopping services. Behavior is not consistent with user-docker/system-docker. I still need to investigate how services are started.

Working in user-docker:
Start a service with `io.rancher.os.remove=false` and container A starts. On each subsequent reboot, container A will be re-started.

Not working in user-docker:
After container A is started, if you make a change to the service, container A is renamed to A-foo and A-foo is started. New container A starts. Essentially A-foo that shouldn't be started as there is no service for it is still started. This happens with any changes made to the service.  I think this is related to #299.

Not working in system-docker:
Start a service with `io.rancher.os.remove=false` and container A starts. On first reboot, container A is renamed to A-foo and A-foo is not started.  New container A starts. Each reboot afterwards, new container A is started as expected since the hash matches. (There were no changes to the service file in the reboots)

docker/container.go
```diff
@@ -47,8 +48,68 @@ func getHash(containerCfg *config.ContainerConfig) (string, error) {
...
+		//var sliceKeys []string
+
+		//Go through keys and write hash
+		for i := 0; i < len(serviceKeys); i++ {
+			serviceValue := unsortedKeyValue[serviceKeys[i]]
+			serviceType := unsortedKeyType[serviceKeys[i]]
+			sliceKeys := []string{}
+
+			//Only need to check for Labels key as it's a Map and write hash after sorting the label names
+			if serviceType == reflect.TypeOf(project.NewSliceorMap(make(map[string]string))) {
```
ibuildthecloud: go has a special syntax for this https://golang.org/doc/effective_go.html#type_switch

docker/container.go
```diff
@@ -47,8 +48,68 @@ func getHash(containerCfg *config.ContainerConfig) (string, error) {
...
+				for lkey := range serviceValue.(*project.SliceorMap).MapParts() {
+					sliceKeys = append(sliceKeys, lkey)
+				}
+				sort.Strings(sliceKeys)
+
+				for j := 0; j < len(sliceKeys); j++ {
+					w.Write([]byte(fmt.Sprintf("%s%v", sliceKeys[j], serviceValue.(*project.SliceorMap).MapParts()[sliceKeys[j]])))
```
ibuildthecloud: can you make the string '%s=%v' so that `ab: cd` hashes different from `a: bcd`

deniseschannon:
@ibuildthecloud Updated based on comments.

PR #302, opened 2015-05-10
rename `rancherctl` to `ros`

imikushin:


deniseschannon:
@imikushin Added a PR to update the docs to correlate with this change.

PR #303, opened 2015-05-11
Update to newer compose

ibuildthecloud:


imikushin:
It actually kinda works!
There are still issues: 
- `FATA[0022] API error (500): Invalid volume specification: /foo` with this compose template: https://gist.github.com/imikushin/2ae44ca5785e93368789
- services with `io.rancher.os.detach=false` dependencies **correctly** only start after such dependencies have finished, BUT they **incorrectly** do start even if such dependencies exit with non-0 status.

PR #305, opened 2015-05-11
Updated for ros instead of rancherctl

deniseschannon:
Whenever #302 is merged, we will want to merge in the docs change. Currently, I have assumed it will be in v0.3.1. If not, I'll need to update.


imikushin:
Wow, that was a big rename!

deniseschannon:
Would we want to document that `rancherctl` still works along with `ros`?

imikushin:
I think it's good to say it's deprecated (and will likely be removed in v0.4.0)

PR #306, opened 2015-05-12
Updated for checking service enable

### LINKED ISSUE #209: rancherctl service enable: when using a custom service, it's annoying that the file must be placed in a specific folder location
It's annoying that I have to save a folder in a specific folder location (`/var/lib/rancher/conf/`) AND I have to name the the service `/var/lib/rancher/conf/example.yml` in order for my service to work.

deniseschannon:
Fixes #209 based on comments about checking for http.


deniseschannon:
This won't allow `ubuntu-console` or any services from os-services to be enabled.

PR #308, opened 2015-05-14
Bugfix: cloud_init datasource "cmdline" should not expect a value

janeczku:
Problem: Datasource of type `cmdline` is not processed unless a value is given. Alas, cloud-init's `proc_cmdline` method does not take any arguments at all.


ibuildthecloud:
Thanks

Issue #313, opened 2015-05-14
Startup Errors: Boot order of containers is not correct


deniseschannon:
After running `sudo mkfs.ext4 -L RANCHER_STATE /dev/sda`, I just continue to reboot.
 
If I continue to reboot over and over again (no changes to OS), you will eventually hit a startup error related to the start of containers.

The startup complains that `all-volumes` container is non-existent and doesn't start a series of other containers.

It also complains that the name "cloud-init-pre-982340825bc9823492349823" is already in use by another container and it needs to be deleted/renamed to be able to resue the container.

Once I get this error, my containers and names are all out of sync and it keeps complaining and will not start all of the system containers.

imikushin:
Confirm reproduced.

deniseschannon:
@imikushin Is this fixed with the container hashing?

imikushin:
Yes

PR #316, opened 2015-05-17
fix container hashing

imikushin:


imikushin:
Just in case: squashed some changes into the last commit (fix getHash()): 
Updated the fix with missing fields: MigrateVolumes, ReloadConfig, CreateOnly. 
Updated TestHash().

imikushin:
also fixes #313

PR #317, opened 2015-05-19
docker image preloading

imikushin:
technology preview ;)

### DIFF
--- a/config/default.go
+++ b/config/default.go
@@ -203,6 +203,22 @@ func NewConfig() *Config {
 				},
 				LogDriver: "json-file",
 			},
+			"preload-system-images": {
+				Image:      "preload",
+				Privileged: true,
+				Labels: project.NewSliceorMap(map[string]string{
+					DETACH:        "false",
+					SCOPE:         SYSTEM,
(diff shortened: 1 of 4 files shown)


imikushin:
This works for both system and user docker. Tested it with modified os-installer: rancherio/os-installer#13

ibuildthecloud:
Can we touch a file when the preload is done?   I only want the preload to run on the first boot, not subsequent.  This is what I'm thinking.  For each file X you touch X.done.  If X has a new timestamp then X.done, you ignore X.done reload.

Also, I really need ISOs to work too.  Do you have a solution for packaging an archive in the ISO?

Finally, can you make "network" service link to preload-system-images.  The reason for this is on start, when loading the compose template, if the image for the service is not found RancherOS automatically adds a link (dependency) to "network".  The idea being that if we have to pull an image, then networking needs to be up.  So as a trick we can ensure that if you preload images plus use the service from your cloud-config, we can ensure the preloading will happen first.  Now this only works with the system docker.  For user-docker I don't know how we can address this.

PR #320, opened 2015-05-24
cloud-init: add support for IP address variable substitution in user-data

janeczku:
This adds support for the cloud-init IP address templating feature.
See https://coreos.com/docs/cluster-management/setup/cloudinit-cloud-config/     
The $public_ipv4, $private_ipv4, etc. placeholders will be substituted in the user-data with the respective values from the meta-data. This happens before saving the config files so that these variables will work in cloud-config, userscript as well as compose style user-data.

**Example use cases in RancherOS:**
* Configure eth1 with a static IP address from `metadata.PrivateIPv4` (rancher config syntax in the user-data file)
* Bind the user-docker API (-H) to the cloud-only IP address (rancher config syntax in the user-data file)
* Bind specific containers to the cloud-only IP address (compose syntax in the user-data file)

### DIFF
--- a/cmd/cloudinit/cloudinit.go
+++ b/cmd/cloudinit/cloudinit.go
@@ -35,6 +35,7 @@ import (
 	"github.com/coreos/coreos-cloudinit/datasource/metadata/ec2"
 	"github.com/coreos/coreos-cloudinit/datasource/proc_cmdline"
 	"github.com/coreos/coreos-cloudinit/datasource/url"
+	"github.com/coreos/coreos-cloudinit/initialize"
 	"github.com/coreos/coreos-cloudinit/pkg"
 	"github.com/coreos/coreos-cloudinit/system"
 	"github.com/rancherio/os/cmd/cloudinit/hostname"
@@ -229,6 +230,7 @@ func saveCloudConfig() error {
 		}
 	}
 
+	userDataBytes = substituteUserDataVars(userDataBytes, metadata)
 	userData := string(userDataBytes)
 	scriptBytes := []byte{}
 
@@ -513,4 +515,11 @@ func toCompose(bytes []byte) ([]byte, error) {
 			"services": compose,
 		},
 	})
+}
+
+func substituteUserDataVars(userDataBytes []byte, metadata datasource.Metadata) []byte {
+	env := initialize.NewEnvironment("", "", "", "", metadata)
+	userData := env.Apply(string(userDataBytes))
+
+	return []byte(userData)
 }
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -23,6 +23,11 @@
 			"Rev": "405c2600b19ae77516c967f8ee8ebde5624d3663"
 		},
 		{
+			"ImportPath": "github.com/coreos/coreos-cloudinit/initialize",
+			"Comment": "v1.3.2-6-g405c260",
+			"Rev": "405c2600b19ae77516c967f8ee8ebde5624d3663"
+		},
+		{
 			"ImportPath": "github.com/coreos/coreos-cloudinit/network",
 			"Comment": "v1.3.2-6-g405c260",
 			"Rev": "405c2600b19ae77516c967f8ee8ebde5624d3663"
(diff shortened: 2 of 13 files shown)


ibuildthecloud:
LGTM

@imikushin Can you try this out?  The code should be sound as it is pulled from CoreOS https://github.com/coreos/coreos-cloudinit/blob/master/initialize/env.go#L76

ibuildthecloud:
Actually, @janeczku Do you just want to use the CoreOS code?  You don't have to use their constructor, just

```go
i := &Environment{substitutions: ..., metadata: ...}
i.Apply(data)
```

janeczku:
@ibuildthecloud Sure can do. It looks like `github.com/coreos/coreos-cloudinit/initialize` was removed from the godeps workdir. Should i put it back in then?

cmd/cloudinit/cloudinit.go
```diff
@@ -230,6 +231,7 @@ func saveCloudConfig() error {
 	}
 
 	userData := string(userDataBytes)
+	userDataBytes = substituteVars(userData, metadata)
```
imikushin: Strings are immutable, so `userData` is left unchanged. It's probably not what you want.

imikushin:
If you import `"github.com/coreos/coreos-cloudinit/initialize"` in cloudinit.go and then do 
```
godep restore
rm -rf Godeps
godep save
```
, you'll have `github.com/coreos/coreos-cloudinit/initialize` added to Godeps.

Explanation:
We use `godep` to manage dependencies. Make sure you have its binary in `$GOPATH/bin` 
(build with `go get github.com/tools/godep`). 

To minimize conflicts, use a dedicated dir as RancherOS GOPATH: e.g. mine is `GOPATH=/home/ivan/src/rancherio/os`.

cmd/cloudinit/cloudinit.go
janeczku: Hi @imikushin, it's exactly what i want though! :) userData does not need to be changed. It is just used for `IsScript` / `IsCompose` and then discarded anyway. It is userDataBytes that needs to be assigned the new value.

janeczku:
> If you import "github.com/coreos/coreos-cloudinit/initialize" in cloudinit.go and then do ...

Thats very helpful, thanks indeed! Godep always gives me headaches...

cmd/cloudinit/cloudinit.go
```diff
@@ -229,7 +230,9 @@ func saveCloudConfig() error {
 		}
 	}
 
-	userData := string(userDataBytes)
+	env := initialize.NewEnvironment("", "", "", "", metadata)
+	userData := env.Apply(string(userDataBytes))
+	userDataBytes = []byte(userData)
```
janeczku: Much less invasive now, just 3 lines. You think it is okay, @imikushin?

imikushin:
Can you also add a unit test? Look at docker/container_test.go for imports and samples. We don't have many of them, but they are obviously needed for changes like this: at least as examples to illustrate what's going on.

cmd/cloudinit/cloudinit.go
imikushin: Just to make myself clear about the unit test I requested: please, turn these lines into a function and add a test illustrating what it does. `saveCloudConfig()` as a whole is, frankly, not unit-testable as is.

janeczku:
@imikushin okay, no problem

cmd/cloudinit/cloudinit.go
janeczku: @imikushin to be clear, you want me to implement them as a function in the unit test file, or you also want them as a function in the cloudinit.go file?

imikushin: as a private function in cloudinit.go: the function would then be called by `saveCloudConfig()` and, of course, by the test.

janeczku:
@imikushin  Added unit test as requested.

```bash
+ go test ./cmd/cloudinit
ok  	_/source/cmd/cloudinit	0.019s
+ exit 0
```

imikushin:
@janeczku Perfect! LGTM 
@ibuildthecloud please merge

PR #338, opened 2015-05-28
empty ServicesInclude map in default config

imikushin:
empty ServicesInclude map in default config

it's not needed anymore as we are using os-services repo


imikushin:
Going to be obsoleted by a next PR

PR #341, opened 2015-05-29
Docs: pre-packing docker images

imikushin:
Docs: pre-packing docker images


deniseschannon:
I merged this, but next time when you do a fenced code block. If you put in 
"```bash" or "```sh"

It'll give the code block a console look. I've updated and pushed a separate change for it.

PR #346, opened 2015-06-01
Fix multiple lines with IP addresses on login screen

### LINKED ISSUE #314: Login screen interface IP addresses are printed multiple times
Sometimes during the boot of RancherOS, it prints out several lines of the IP information made from this change (https://github.com/rancherio/os/pull/247).

Steps to Reproduce:
Reboot several times after saving to disk or setting in persisting state.

In error #313, I got this:
```
docker-sys: 172.18.42.1 docker0: 172.17.42.1 eth0: 10.0.2.15 lo: 127.0.0.1
docker-sys: 172.18.42.1 docker0: 172.17.42.1 eth0: 10.0.2.15 lo: 127.0.0.1
docker-sys: 172.18.42.1 eth0: 10.0.2.15 lo: 127.0.0.1
```

Other times: 
```
docker-sys: 172.18.42.1 docker0: 172.17.42.1 eth0: 10.0.2.15 lo: 127.0.0.1
docker-sys: 172.18.42.1 docker0: 172.17.42.1 eth0: 10.0.2.15 lo: 127.0.0.1
```

janeczku:
fixes #314: Login screen IP addresses are printed multiple times

Signed-off-by: Jan Broer <janeczku@yahoo.de>

### DIFF
--- a/scripts/dockerimages/scripts/console.sh
+++ b/scripts/dockerimages/scripts/console.sh
@@ -104,6 +104,7 @@ if ! grep -q "$(hostname)" /etc/hosts; then
     echo 127.0.1.1 $(hostname) >> /etc/hosts
 fi
 
+echo 'RancherOS \n \l' > /etc/issue
 echo $(/sbin/ifconfig | grep -B1 "inet addr" |awk '{ if ( $1 == "inet" ) { print $2 } else if ( $2 == "Link" ) { printf "%s:" ,$1 } }' |awk -F: '{ print $1 ": " $3}') >> /etc/issue
 
 if [ -x /opt/rancher/bin/start.sh ]; then


ibuildthecloud:
What is odd to me is that the console should be ephemeral so you shouldn't need to reset `/etc/issue`.  This issue would happen if you restarted the console, but rebooting RancherOS should delete the console.  I wonder if there is a bug there.

janeczku:
@ibuildthecloud  My suspicion is that this issue occurs when you enable the `ubuntu-console` service.  `/usr/sbin/console.sh` would then run twice (as it's invoked by both the `06-console` and `ubuntu-console` containers).

ibuildthecloud:
@janeczku We've greatly refactored the build and will include this fix in 0.4 but it needs to go to rancher/os-images.  We will keep this PR open until we port this over to rancher/os-images.

janeczku:
@ibuildthecloud Allright :+1:

imikushin:
Continuing as rancher/os-images#15

PR #347, opened 2015-06-01
add ClientAliveInterval 180 to sshd_config

imikushin:
add ClientAliveInterval 180 to sshd_config


imikushin:
Should prevent SSH session from timing out by sending null packets to client every 3 minutes. Required by MS Azure certification tests.

PR #354, opened 2015-06-05
make release files available for download

imikushin:
make release files available for download

both `http:` and `https:` are supported


ibuildthecloud:
Can you keep the original section for ISO but change the URL to GCS.  Then after than add a section for `Additional Downloads`.  That way it is still obvious what to download.  So basically

### ISO

https://releases..../v0.3.1/rancheros.iso

### Additional Download

...all URLs...

imikushin:
Done

PR #361, opened 2015-06-09
make /var/lib/rancher persistent

imikushin:
make /var/lib/rancher persistent

by making it a system volume


ibuildthecloud:
Can we just do /var/lib/rancher/state?  I swear there is a bug somewhere if we do the full directory.  I wish I could remember what it was.

imikushin:
Can do (and, no problem, will do if you insist). 
Just a proposal: should we maybe let the bug show its ugly face and fix it properly, with a test?

PR #374, opened 2015-06-11
refactor installing rancheros to disk docs

wlan0:
@sheng-liang @deniseschannon

### DIFF
--- a/docs/cloud-config/index.md
+++ b/docs/cloud-config/index.md
@@ -87,3 +87,12 @@ rancher:
         - 8.8.4.4
 
 ```
+
+#### Set Hostname
+
+You can set the hostname of the host using cloud-config. The example below shows how to configure it.
+
+```yaml
+hostname: myhost
+```
+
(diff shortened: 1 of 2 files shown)


docs/cloud-config/index.md
```diff
@@ -87,3 +87,12 @@ rancher:
...
 
 ```
+
+#### Set Hostname
+
+You can set the hostname of the host using cloud-config. The example below shows how to configure it.
+
+```yaml
+hostname: myhost
+```
```
deniseschannon: Can you make notes about how if you set net: host, then the hostname will automatically be "rancher"? We probably need a better explanation than what I have put in the issue, but I don't understand why it gets overwritten.

https://github.com/rancherio/os/issues/217

deniseschannon: Never mind on my comment based on our discussion that this is how to overwrite the "rancher" hostname for RancherOS.

PR #391, opened 2015-06-17
Fix typo in ros export command

radriaanse:
Should be ros config export -p, not ros export config -p.


deniseschannon:
@radriaanse Thanks for the fix!

PR #392, opened 2015-06-17
Refactored build

imikushin:
Pulls only needed system services images from dockerhub based on os-config.yml

Do not merge yet: needs to be rebased when rancherio/os#386 is merged.


imikushin:
@ibuildthecloud Please review /cc @deniseschannon 

I'm pretty sure this is good to merge though ;)

scripts/run
```diff
@@ -48,31 +48,25 @@ if [ ! -d ${INITRD_TMP} ]; then
...
+cp dist/artifacts/rancheros ${INITRD_TMP}/init
+cp -f os-config.yml ${INITRD_TMP}/
 cd ${INITRD_TMP}
 
 find | cpio -H newc -o > ${INITRD_TEST}
 
 if [ ! -e ${HD} ]; then
     mkdir -p $(dirname ${HD})
-    zcat ${HD_GZ} > ${HD}
+    gzcat ${HD_GZ} > ${HD}
```
ibuildthecloud: gzcat doesn't exist on linux

ibuildthecloud: If you want a cross platform way of doing this you can do `gzip -dc ${HD_GZ} > ${HD}`

scripts/build-images
```diff
@@ -1,46 +1,47 @@
...
-    docker build -t $tag -f $i .
-    docker build -t ${tag}:${VERSION} -f $i .
-    echo $tag >> ${BUILD}/tags
-done
+ln -sf bin/rancheros ./ros
 
-echo Creating images.tar
-docker save $(<${BUILD}/tags) > ${BUILD}/initrd/images.tar
+for i in $(./ros c images -i os-config.yml); do
+    docker pull ${i}:${VERSION}
```
ibuildthecloud: This seems odd that the version is taken from here.  I would rather the version be in the `os-config.yml`.  I know it's a pain to update that file, but then we make is such that every single image in the `os-config.yml` does not need to be bumped with each ros release.

scripts/build-images
```diff
@@ -1,46 +1,47 @@
...
+    docker tag -f ${i}:${VERSION} ${i}
```
ibuildthecloud: Why do you need to tag after the pull?

ibuildthecloud: Oh I see now.  It really would be better if we have the version in the os-config.yml

imikushin: Yeah, the script will become a bit simpler then.

scripts/build-images
imikushin: I did have that feeling for a moment... Now I see it was right: because to be able to really pull everything together from this file we need full image names, so versions must be there. Will fix.

scripts/run
imikushin: yeah, and on osx zcat wants the source file to have .Z suffix so much it imagines it's there (even if it's not) and complains empty-hd.img.gz.Z doesn't exist! 

thanks for pointing this out.

PR #428, opened 2015-07-05
Disable userland proxy for user-docker and system-docker

Hellslicer:
Since Docker 1.7, we can disable the userland proxy (it remains enabled by default due to old kernels).

docker/docker#12165 / docker/libnetwork#171


ibuildthecloud:
@Hellslicer Can you make the corresponding change for system_docker configuration in the same file?

Hellslicer:
Done.
I did not believe it was useful as the system containers use the parameter `--net=host`.

ibuildthecloud:
Thanks @Hellslicer

PR #441, opened 2015-07-14
really fix the rancher/dind:v0.5.0 regression

imikushin:
really fix the rancher/dind:v0.5.0 regression

not just avoid it


ibuildthecloud:
LGTM

PR #450, opened 2015-07-23
dind-less, make-based build system

imikushin:
WIP: make-based build system


.drone.yml
```diff
@@ -1,3 +1,3 @@
 image: rancher/dind:v0.1.0
 script:
- - ./scripts/ci
+ - ./build.sh
```
ibuildthecloud: Should this point to ./scripts/ci  and ./scripts/ci should call ./scripts/bootstrap and ./build.sh.  Bootstrap should `apt-get install make` and call wrapdocker if available

imikushin:
@ibuildthecloud Looks like it's ready

ibuildthecloud:
My general comment is that there is a lot of logic in the makefile now.  As opposed to the make file being a very simple wrapper to shell scripts.  I'll review it in depth a bit later.

PR #456, opened 2015-07-28
do not pull os-images by default if they exist

imikushin:
do not pull os-images by default if they exist

also, cleanup os-initrdbase container on exit in ./scripts/copy-images
(as trap can apparently only run one command)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@
 DOCKER_BINARY_URL := https://github.com/rancher/docker/releases/download/v1.7.1-ros-1/docker-1.7.1
 
 
+FORCE_PULL   := 0
 DOCKER_BINARY := $(shell basename $(DOCKER_BINARY_URL))
 pwd := $(shell pwd)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -56,7 +57,7 @@ INITRD_DIR := $(BUILD)/initrd
 
 $(INITRD_DIR)/images.tar: bin/rancheros
 	ln -sf bin/rancheros ./ros
-	for i in `./ros c images -i os-config.yml`; do docker pull $$i; done
+	for i in `./ros c images -i os-config.yml`; do [ "$(PULL_IMAGES)" != "1" ] && docker inspect $$i >/dev/null 2>&1 || docker pull $$i; done
```
ibuildthecloud: One line is not greater than 3 :)

ibuildthecloud: Honestly this line is really hard to read.  You have a negated condition followed by a && then ||

PR #460, opened 2015-07-31
make cloud-config the configuration mechanism for RancherOS

imikushin:
Do not merge yet. 

A very obscure bug (and now traced) was causing `ros config get` to 
print out YAML unmarshaling errors instead of configuration values.
`c.readArgs()` usage is now commented out (config/config.ro:228), and I need to 
code the proper fix now. 

Currently, it finally compiles and runs, but needs cleanup.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -2,58 +2,21 @@ package config
 
 import (
 	"io/ioutil"
-	"os"
 	"strings"
-
-	"github.com/rancherio/rancher-compose/librcompose/project"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/rancherio/os/util"
+	"github.com/rancherio/rancher-compose/librcompose/project"
 	"gopkg.in/yaml.v2"
 )
 
-func (c *Config) privilegedMerge(newConfig Config) error {
-	err := c.overlay(newConfig)
-	if err != nil {
-		return err
-	}
-
-	for k, v := range newConfig.SystemContainers {
-		c.SystemContainers[k] = v
-	}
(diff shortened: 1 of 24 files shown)


config/config.go
```diff
@@ -12,27 +11,27 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
-func (c *Config) privilegedMerge(newConfig Config) error {
+func (c *RCloudConfig) privilegedMerge(newConfig RCloudConfig) error {
```
ibuildthecloud: Rename CloudConfig

ibuildthecloud:
Lets delete `system_container` from config and just use `services`.  Also remove any "readonly" logic in config.go

config/types.go
```diff
@@ -24,14 +27,17 @@ const (
 	RELOAD_CONFIG = "io.rancher.os.reloadconfig"
 	SCOPE         = "io.rancher.os.scope"
 	SYSTEM        = "system"
+
+	OsConfigFile          = "/os-config.yml"
+	CloudConfigFile       = "/var/lib/rancher/conf/cloud-config.yml"
+	CloudConfigScriptFile = "/var/lib/rancher/conf/cloud-config-script"
+	MetaDataFile          = "/var/lib/rancher/conf/metadata"
+	LocalConfigFile       = "/var/lib/rancher/conf/local-config.yml"
```
ibuildthecloud: rename to cloud-config-local.yml

imikushin:
fixes #454

PR #463, opened 2015-08-05
simpler make

imikushin:
divide Makefile into local / in-docker. 
Move chunks of shell to scripts.


ibuildthecloud:
Now that looks like a sane makefile

Issue #464, opened 2015-08-05
Calling `ros env` without arguments results in Panic: Runtime Error


janeczku:
**Steps to reproduce**
In RancherOS v0.3.3 type `sudo ros env`

**Expected result**
Command returns no error

**Actual result**
```
panic: runtime error: index out of range

goroutine 1 [running]:
github.com/rancherio/os/cmd/control.envAction(0xc20808c0c0)
	/source/gopath/src/github.com/rancherio/os/cmd/control/env.go:31 +0x6fb
github.com/codegangsta/cli.Command.Run(0xa28090, 0x3, 0xa27d10, 0x1, 0xa477d0, 0xb, 0x0, 0x0, 0x0, 0x0, ...)
	/source/Godeps/_workspace/src/github.com/codegangsta/cli/command.go:116 +0x104b
github.com/codegangsta/cli.(*App).Run(0xc20800c0f0, 0xc20800a000, 0x2, 0x2, 0x0, 0x0)
	/source/Godeps/_workspace/src/github.com/codegangsta/cli/app.go:172 +0xdf6
github.com/rancherio/os/cmd/control.Main()
	/source/gopath/src/github.com/rancherio/os/cmd/control/cli.go:62 +0x34d
github.com/docker/docker/pkg/reexec.Init(0xa8c570)
	/source/Godeps/_workspace/src/github.com/docker/docker/pkg/reexec/reexec.go:26 +0x71
main.main()
	/source/main.go:53 +0x20d
```

PR #465, opened 2015-08-05
Return gracefully when `ros env` is called without argument(s)

janeczku:
Fixes Runtime error issue #464


ibuildthecloud:
Thanks!

PR #468, opened 2015-08-06
libcompose-refactor: fix user-docker

imikushin:


imikushin:
Supersedes #461

ibuildthecloud:
Was there a technical reason for renaming the udev services?  Or just for clarity?

imikushin:
I renamed `udev`s because there were naming conflict messages during boot. Also, for clarity.

PR #479, opened 2015-08-12
update to docker-1.8.1

imikushin:
update to docker-1.8.0


imikushin:
Tests should pass when `rancher/docker:1.8.0` is available on DockerHub (see rancher/docker-from-scratch#11)

PR #481, opened 2015-08-12
WIP: fix services_include

imikushin:
fix services_include


imikushin:
WIP because there's a fix that should really go to libcompose first.

PR #355, opened 2015-06-06
Logrotate. Issue #329

konstruktoid:
Daily rotation of /var/log/messages and /var/log/*docker*
Issue  #329 "Rotate /var/log/messages"

The indentation is wrong, so could you tell which settings I should use?

Signed-off-by: Thomas Sjögren <konstruktoid@users.noreply.github.com>

PR #484, opened 2015-08-13
WIP: add logrotate

imikushin:
add logrotate

porting rancher/os#355
completes rancher/os-images#16

### DIFF
--- a/os-config.yml
+++ b/os-config.yml
@@ -157,6 +157,18 @@ rancher:
       read_only: true
       volumes:
       - /var/lib/docker:/var/lib/docker
+    logrotate:
+      image: rancher/os-logrotate:v0.4.0-dev
+      labels:
+        io.rancher.os.scope: system
+      log_driver: json-file
+      net: none
+      pid: host
+      ipc: host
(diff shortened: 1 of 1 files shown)


imikushin:
Do not yet merge: native docker rollover logging might be a better option. 

Also, can only be merged, when rancher/os-logrotate:v0.4.0-dev is in dockerhub.

ibuildthecloud:
@imikushin I think for container logging we should switch to rollover driver.  Can you set it up to rollover after 50mb and keep only one 1 log (no history)?  The logrotate is for something different though, `/var/log/*docker.log` is for the daemon logs and not the container logs.  This doesn't seem quite right to do `restart: always` if logrotate isn't a daemon.  We need a different approach for cron like jobs.  Maybe we should have a cron container and share /etc/crontab + the system-volumes with the container.  Then users can do cron.

imikushin:
@ibuildthecloud I misunderstood that comment (https://github.com/rancher/os/pull/484#issuecomment-130854339). I restored the line that was missing.

PR #505, opened 2015-08-17
fix boot logging on `rancheros-install` installed systems

imikushin:
Before this fix we wouldn't see what services have started or failed. 
System init messages not appearing in this log were making RancherOS 
look like it silently hangs on boot.


imikushin:
fixes #502

Issue #501, opened 2015-08-17
v0.4.0-rc1: rancheros-install installs an unbootable system


imikushin:
Crash when booting a rancheros-install installed system.

imikushin:
Fixed by #506

PR #506, opened 2015-08-17
fix the boot crash on installed system

imikushin:
The crash is described in #501

### DIFF
--- a/os-config.yml
+++ b/os-config.yml
@@ -37,7 +37,7 @@ rancher:
       - /lib/modules:/lib/modules
       - /lib/firmware:/lib/firmware
   bootstrap_docker:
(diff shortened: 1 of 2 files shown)


imikushin:
Here's the serial console log booting the system installed from this branch 

https://gist.github.com/imikushin/9552993053d01cbd71ae

os-config.yml
```diff
@@ -37,7 +37,7 @@ rancher:
       - /lib/modules:/lib/modules
       - /lib/firmware:/lib/firmware
   bootstrap_docker:
-    args: [-d, -s, overlay, -b, none, --restart=false, -g, /var/lib/system-docker,
```
ibuildthecloud: We need --restart=false.  I know it's deprecated but we need this.  This is so that autorestart containers do start on system start.

Issue #487, opened 2015-08-13
v0.4: Invalid cloud-config panic boots


ibuildthecloud:
If you pass in a poorly formatted cloud-config `./script/run --cloud-config bad.yml` on boot it panics.  Couple things.  We should never show a panic unless some really bad happened.  I know a panic shows a stack traces but a good error message is just as useful for debugging.  Also we shouldn't halt the boot if the cloud-config is bad.

Finally, it seems that once you pass in a bad cloud-config you can never boot the system again as it has cached the file.

imikushin:
should be fixed now ;)

PR #512, opened 2015-08-20
less panics and fatals

imikushin:
Don't throw panics where logging fatal or error would do.
Don't save bad cloud-config.
Don't abort cloud-init on configuration errors, only report them.


imikushin:
This should fix #487

ibuildthecloud:
LGTM

Issue #498, opened 2015-08-14
v0.4.0-rc1: ros config export prints out information in rancher key on fresh install


deniseschannon:
I'd expect that `ros config export` to print out only what is in the cloud-config-local.yml.

With a fresh iso install and no changes, if i run the command, it shows in the rancher key.

```
rancher:
  state:
    autoformat: 
      - /dev/sda
      - /dev/vda
```

I would expect that it would be blank like the cloud-config-local.yml file is 
```
sudo cat /var/lib/rancher/conf/cloud-config-local.yml
{}
```

ibuildthecloud:
This is because the ISO auto sets these parameters.  @imikushin Looks like there is a merging issue in that we have two `/dev/sda` values there.

deniseschannon:
There was a typo in my original post, but @imikushin to check if append the same item on two lists will not show up duplicates.

imikushin:
@deniseschannon `ros config export` also outputs configuration values set by `rancher.` prefixed kernel parameters

deniseschannon:
The changes in the file is due to loading from the ISO.

PR #514, opened 2015-08-21
slices union and intersection

imikushin:
slices union and intersection

useful for merging and splitting configs


imikushin:
This should fix potential configuration file merging issues (mentioned in #498)

PR #519, opened 2015-08-22
Fix sha1sum call to work on both Linux and Darwin.

bruienne:
When running `scripts/run` on Darwin the xhyve boot process fails due to a missing root device. This is because `sha1sum` is called to generate a temp folder using the SHA1 hash of `dist/artifacts/initrd`  as its name. This temp folder is then used to generate the Rancher OS `initrd.test` to boot xhyve. Since `sha1sum` is named `shasum` on Darwin the command fails and `$INITRD_TMP` is set to `./build` instead and the generated `initrd.test` file is unbootable. This PR fixes the issue by setting `sha1sum` to the result of `which sha1sum` for Linux hosts and the result of `which shasum` for Darwin hosts. Tested and working as intended on a clean 10.10.5 host.

### DIFF
--- a/scripts/run
+++ b/scripts/run
@@ -7,16 +7,21 @@ cd $(dirname $0)/..
 source scripts/build-common
 
 BASE=$(pwd)
+UNAME=$(uname)
+
+# Linux and Darwin SHA1 sum binary are different, pick which to use
+if [ "$UNAME" == "Darwin" ]; then sha1sum=$(which shasum)
+elif [ "$UNAME" == "Linux" ]; then sha1sum=$(which sha1sum);
+fi
 
 KERNEL=${BASE}/dist/artifacts/vmlinuz
 INITRD=${BASE}/dist/artifacts/initrd
 NO_COMPRESS_INITRD=${INITRD}.none
 HD=${BASE}/state/empty-hd.img
 HD_GZ=${BASE}/assets/empty-hd.img.gz
-INITRD_TMP=${BUILD}/$(sha1sum ${INITRD} | awk '{print $1}')
+INITRD_TMP=${BUILD}/$(${sha1sum} ${INITRD} | awk '{print $1}')
 INITRD_CURRENT=${BUILD}/initrd-current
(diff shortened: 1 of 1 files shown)


imikushin:
@bruienne when I switched to OS X, I also hit this, but quickly found an alternative: 
```
brew install md5sha1sum
```
This installs `md5sum` and `sha1sum` binaries and makes the problem go away :)

Seriously though, I think I should write a small blurb in the README on how to start developing w/ OS X.

imikushin:
@bruienne BTW, you also need `coreutils` (for `greadlink`), `cdrtools` (for `mkisofs`) and `qemu` (for `qemu-img`) installed to be able to successfully run `./scripts/run`

bruienne:
@imikushin I did also install the required tools with brew, but since I've run into the `sha1sum` vs. `shasum` issue in the past I didn't think it required installing overlapping tools. Obviously the tools in `coreutils`, `cdrtools` (actually that needed `dvdrtools` for me) and `qemu` are not built in to Darwin so it was required to install them. You're right that it's more of a documentation issue than code. Thought I'd offer the PR with the fix in case it would be useful to the project.

ibuildthecloud:
@bruienne Does `shasum -a 1` work across both.  I think I've ran into this before and found that using `shasum` and passing the algorithm was the most compatible incarnation.  (fyi, Not a Mac user, don't have a Mac readily available)

bruienne:
@ibuildthecloud It doesn't - sadly. The `sha<algo>sum` commands on Linux are not aliases but actual separate binaries so there's no multi-purpose `shasum` to call. One other solution could be to use `openssl` instead to generate the hash: `openssl dgst -sha1 <filename>` - that's pretty certain to be available across platforms. It comes with OS X by default. If you'd like I can update the PR using `openssl` instead.

ibuildthecloud:
Oh bummer.  I run Ubuntu/Debian and I do have `shasum` but maybe that isn't consistent across all Linux distros.

PR #529, opened 2015-08-27
Storage

ibuildthecloud:


ibuildthecloud:
I'll update name to `user-docker`

imikushin:
Reverted: merging it broke console (although by itself, the branch builds and runs fine).

These last few lines of the broken boot log might be of interest: 
https://gist.github.com/imikushin/4e9bf39118dd168a6941

PR #547, opened 2015-09-02
Revive integration tests

imikushin:
Revive integration tests in Python 2.7. 
Works on Linux (with and without KVM) and on OS X.


imikushin:
@ibuildthecloud The tests have been successfully run by Drone and passed!

Issue #548, opened 2015-09-02
Lots of messages appear on the console


sheng-liang:
I installed 4.0 RC5 to /dev/sda, and set grub.conf to enable console login. After reboot and login to the console, I see lots of messages on the console. They seem to be network related. There's an error about duplicate IPv6 address.
![messages](https://cloud.githubusercontent.com/assets/9485278/9641464/ad9fa56e-516b-11e5-992e-7de94594b909.png)

imikushin:
These messages are kernel logging. They can be disabled with a kernel parameter `quiet` in the bootloader config. I'd leave them on right now, but this is really a matter of preference.

sheng-liang:
I did not see them in prior releases. Something changed? Shouldn't the messages go into a log?

deniseschannon:
When booting from iso, we have made the kernel logging quiet, but by default, we have not turned on `quiet` for when installing to disk. 

We will not be changing this behavior.

PR #549, opened 2015-09-03
Quiet down kernel logging

imikushin:
Quiet down kernel logging so that it doesn't flood user consoles as in #548


ibuildthecloud:
Did we previously have quiet on?

imikushin:
No

imikushin:
But we've had it on in the ISO.

Issue #520, opened 2015-08-23
32-bit build issue: Reboot takes int but LINUX_REBOOT_CMD_HALT overflows int


luxas:
Tries now to build bin/rancheros like in Makefile.docker on a 32 bit machine (a Raspberry Pi, in fact)

The build stopped in cmd/power, on the 129th line.

It is this issue: golang/go#9584

I will patch this for my RPi

luxas:
Closing this because its merged

PR #559, opened 2015-09-10
Fixed 32-bit issue in power.go

luxas:
Hello,

Here is a fix to #520.

I haven't tested it on a 64-bit machine, but you may know if it is save to merge.

I would think so, just a small conversion.

Another possibility is to make a 32bit experiment branch, which I would prefer.

### DIFF
--- a/cmd/power/power.go
+++ b/cmd/power/power.go
@@ -129,7 +129,7 @@ func Halt() {
 	reboot(syscall.LINUX_REBOOT_CMD_HALT)
 }
 
-func reboot(code int) {
+func reboot(code uint) {
 	err := shutDownContainers()
 	if err != nil {
 		log.Error(err)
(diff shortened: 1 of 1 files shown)


imikushin:
@luxas Thanks for the patch!
It was a compile error on RPi, right?
LGTM
/cc @ibuildthecloud

luxas:
Yes, with that patch its possible to build `bin/rancheros` on the Pi.

What is the status of the RPi 2 issue, btw?
@imikushin

imikushin:
@luxas We're currently wrapping up things for v0.4.0 release, so I haven't had time for RPi (sorry!). However, v0.4.0 (the current master and the v0.4.0 RCs) is much more porting friendly, than the previous versions, so RPi port is now much easier to produce. I'm hoping someone is going to try :)

Issue #573, opened 2015-09-22
Restructure cloud-config.yml to eliminate multiple files and make it simple


deniseschannon:
cloud-config.yml is able to be edited using `ros config`
cloud-config.d is used for user-data and default settings

PR #580, opened 2015-09-24
reshuffle cloud-config

imikushin:
read files cloud-config.d in alphanumeric order, 
then cloud-config-local.yml


ibuildthecloud:
I can't take it, I really want `/var/lib/rancher/conf/cloud-config-local.yml` to be `/var/lib/rancher/conf/cloud-config.yml`.  I think the fact that we call the downloaded one `/var/lib/rancher/conf/cloud-config.d/cloud-config-boot.yml` is enough to disguish between the two.  Also the pattern of `conf-file` and `conf-file.d` is well established.

cmd/cloudinit/cloudinit.go
```diff
@@ -204,49 +124,67 @@ func saveCloudConfig() error {
...
+	if err != nil {
+		log.Errorf("Failed fetching meta-data from datasource: %v", err)
+		return nil, metadata, err
+	}
+	return userDataBytes, metadata, nil
+}
+
 func getSaveCloudConfig() (*config.CloudConfig, error) {
-	ds := file.NewDatasource(rancherConfig.CloudConfigFile)
+	ds := file.NewDatasource(rancherConfig.CloudConfigBootFile)
```
ibuildthecloud: I don't think this works right.  Before, a bit back, we used to merge everything to one file called cloud-config-processed.yml and that was read.  Now I believe we need to just load the default RancherOS config and used that as our cloud-config.  We don't need data sources at all now because we already have the data structures saved to disk.

ibuildthecloud:
This immutability seems really half baked I don't see the value of the change.  Set, Import, Merge, SetConfig all save to disk.  So while the data structure is immutable the code still has a massive side effect which is mutating state on disk.  I think this almost makes it work.  I fell like the old methods were obviously mutating, but now you give the impression of immutability but it is in fact still mutating and a silent and dangerous way.

ibuildthecloud:
Just to be clear, I am all for immutability, but I feel this change is misleading.  I would not expect Set() to save to disk because it returns a CloudConfig.  I would expect a Save() method.  So

```go
cfg, err = cfg.Set(x,y)
if err != nil {
  return err
}

return cfg.Save()
```

compose/project.go
```diff
@@ -80,12 +80,13 @@ func newProject(name string, cfg *config.CloudConfig) (*project.Project, error)
 	return docker.NewProject(context)
 }
 
-func addServices(p *project.Project, cfg *config.CloudConfig, enabled map[string]string, configs map[string]*project.ServiceConfig) {
```
ibuildthecloud: Why??? Why is this changed.

ibuildthecloud:
I don't understand: https://github.com/rancher/os/commit/4f6f87a16cfe57fd8cbc571314c38a5d4fecba57.  You need some tangible benefit to changing the code.  The answer can't be because immutability is better.  In most cases that is a stylistic choice.  The fact that you change `map[string]string` to `map[interface{}]interface{}` makes me feel like there is a bigger impact to this change than I understand.  Which then leads me to why?

When we refactor code because we are adding features, I'm all for changing the style.  I just don't get this change.

ibuildthecloud:
Oh I think I get why you changed, it's because of the your `util.MapCopy` method.  Please revert unless you really have a great reason to change this.

ibuildthecloud:
Can you add some tests before we merge.

cmd/control/tlsconf.go
```diff
@@ -65,16 +65,21 @@ func writeCerts(generateServer bool, hostname []string, cfg *config.CloudConfig,
...
 		}
 
-		return cfg.SetConfig(&config.CloudConfig{
-			Rancher: config.RancherConfig{
-				Docker: config.DockerConfig{
-					CAKey:      cfg.Rancher.Docker.CAKey,
-					CACert:     cfg.Rancher.Docker.CACert,
-					ServerCert: string(cert),
-					ServerKey:  string(key),
+		cfg, err = cfg.Merge(map[interface{}]interface{}{
```
ibuildthecloud: Why can't this be done in a type safe way?  Should we add a `MergeCloudConfig(config.CloudConfig)` method?  I just don't like all these hard coded strings.  And move then to const doesn't make it better.  I'd rather it be type safe.

config/disk.go
```diff
@@ -4,10 +4,24 @@ import (
...
 	"os"
 
+	"github.com/coreos/coreos-cloudinit/datasource"
+	"github.com/coreos/coreos-cloudinit/initialize"
 	"github.com/rancherio/os/util"
 	"gopkg.in/yaml.v2"
 )
 
+var (
+	metadata        datasource.Metadata
```
ibuildthecloud: Let's not cache, this has some odd side effects

imikushin:
I've addressed the comments. Merging and cutting the next RC.

deniseschannon:
https://github.com/rancher/os/issues/573

PR #590, opened 2015-10-01
Support kernels from local directories, but just remote URLs

ibuildthecloud:


maxfierke:
:+1: Oh my! Thank you :D

ibuildthecloud:
@maxfierke If you are looking to use this you must put your kernel tgz in a subdir of the git repo and then set `COMPILED_KERNEL_URL` in `build.conf` to the relative path.  This makes it work in the crazy docker build setup we have.  For example

Create `./assets/kernel.tar.gz`
Set `COMPILED_KERNEL_URL=./assets/kernel.tar.gz`

PR #597, opened 2015-10-06
remove boot behaviour dependency on version string

imikushin:
Remove dependency on version string.

Making version string a part of configuration kinda smells and makes testing stateful behaviour difficult.


imikushin:
I've added a kernel param `rancher.rm_usr` to wipe out the `/usr-{version}` dir on boot and made this the default in `./scripts/run` (can be disabled with `--no-rm-usr`)

PR #598, opened 2015-10-06
WIP: Show logs with no previous tail

imikushin:
WIP: Show logs with no previous tail

So that service logs from previous boots do not appear in the current boot logs.
WIP, Godeps update needed.


imikushin:
Addressed by #659

PR #605, opened 2015-10-13
Replace rancherio with rancher

datawolf:


ibuildthecloud:
Cool thanks! @imikushin Can you check this out.  Would be great to move away from `rancherio/os` we've been slowly doing it with all our other projects.

imikushin:
LGTM

PR #621, opened 2015-10-28
Delete the unused const variable in docker/client.go

datawolf:
The commit 9d76b79ac36e87965 refactor code for function
"NewClient", and we should delete the unused const
variable.


imikushin:
LGTM

PR #622, opened 2015-10-29
docker-1.9

imikushin:
DON'T MERGE. Waiting for docker-1.9.1

### DIFF
--- a/Dockerfile.build-base
+++ b/Dockerfile.build-base
@@ -8,7 +8,7 @@ RUN locale-gen en_US.UTF-8
 ENV LANG en_US.UTF-8
 ENV LC_ALL en_US.UTF-8
 ENV TERM xterm
-RUN curl -sL https://get.docker.com/builds/Linux/x86_64/docker-1.8.2 > /usr/bin/docker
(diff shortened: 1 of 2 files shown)


ibuildthecloud:
@mailq Unfortunately you can't use 1.9 to build RancherOS that has 1.9 in it :disappointed:   https://github.com/docker/docker/issues/17290  There is a bug in 1.9.0 that fails for us.

ibuildthecloud:
@imikushin I'm not so confident on 1.9 quite yet.  I'm trying to find out when 1.9.1 will be out.

imikushin:
@mailq @ibuildthecloud That bug is pretty nasty. Otherwise, I'd use 1.9.0 instead of 1.8.3 in Dockerfile.build-base if it were not for the following problem: 

With docker 1.9 in our build container, the CI build fails on Drone, which apparently has Docker 1.8.
This is happening, because we do not run Docker-in-Docker. In our build container we're using the outside docker server by bind-mounting /var/run/docker.sock

imikushin:
I'd also like to wait for 1.9.1. Let's hold it off.

imikushin:
Mergin' it so we can cut v0.4.1 RC with docker 1.9

PR #625, opened 2015-10-31
config/types.go: add "" to field tag

datawolf:
The field tag format accepted is[1]:
    `(...) yaml:"[<key>][,<flag1>[,<flag2>]]" (...)`
so we should add double quotation marks for the
filed tag format.

[1]: https://godoc.org/gopkg.in/yaml.v2


ibuildthecloud:
@datawolf Feel free to delete `ContainerConfig`, I don't think that type is used anymore.

datawolf:
ok, I will add another patch to delete this type

datawolf:
@ibuildthecloud, `ContainerConfig` have been deleted.

imikushin:
LGTM

Issue #627, opened 2015-11-01
Rancher OS does not respect hostname in network configuration


mailq:
Using latest rancher OS 0.4.0 on bare metal (using iPXE) and a provided cloud-config.yml does not respect the given hostname for DHCP configuration.
In the cloud-config I provide the individual hostname, which gets picked up by rancher as expected. I can see it on the login prompt and also in /etc/hostname. As I leave out the network interface configuration, rancher configures the network via DHCP.
But while issuing the DHCP query, rancher os keeps promoting the hostname "rancher" and not the configured one. I can see that in my dnsmasq logs "client provides name: rancher". So my DHCP/DNS server accepts that name and the rancher box is then only visible on the network via "rancher". You could guess that one can only have one rancher box this way.
So rancher OS must use the hostname when doing DHCP requests.

After some more investigation this is understandable:
The boot-process first starts **[10/17]** the "cloud-init-pre" then **[11/17]** the "network" and thereafter **[12/17]** the "cloud-init". As the hostname is in the cloud-config, but the network is configured before, the hostname can not be used in DHCP requests. If there is no reason for this boot order, one should swap step 11 and 12.

The workaround for this bug is simply to issue an `sudo system-docker restart network` right after booting is finished. (~~One could also put this into the startup script~~ <- no you can't. The script is executed in the console container with no access to the system-docker). This is an ugly workaround as it needs another DHCP request to work, but at least it does the job.

faddat:
I'm also experiencing this behavior, and it's not taking mac addresses I've assigned in cloud-config.

ibuildthecloud:
@mailq I can easily fix the issue in which the hostname is not set before running DHCP.

@faddat I don't believe we set the MAC address.  We use the MAC address to match the configuration.

imikushin:
This should be fixed in v0.4.2-rc2. @mailq @faddat Can you confirm? 

I tested by booting a RancherOS v0.4.2-dev VM (on VirtualBox) on a bridged NIC on a local WiFi network and pinging it by its hostname.

azbpa:
This is quite old issue but facing the same in `v1.5.0`.

I am deploying RancherOS from Rancher vSphere provider, which sets the `hostname` according to node name defined in Rancher UI. 

In general, the hostname gets applied correctly (I can see it on the console and it gets reported correctly by VMware Tools).

However, in the DHCP request it contains the default hostname `rancher` so it gets registered with this name in DNS (what is not intended).

[...]

PR #633, opened 2015-11-04
Set hostname prior to running DHCP

ibuildthecloud:
https://github.com/rancher/os/issues/627


imikushin:
Besides there's no test case, LGTM. 

This is good to merge, but the bug (#627) should stay open until there's a test for it. We can add it to validation tests (TBD)

mailq:
I confirm that this solves #627. Manually tested and approved. I'm unable to write an automated test for that.

PR #634, opened 2015-11-05
system-docker: replace /usr/bin/docker with DOCKER_BIN

datawolf:
It is better to use const DOCKER_BIN.


imikushin:
@datawolf Thanks for submitting this PR: it is indeed a better idea to use a constant instead of a "magic" value. Although, only the bundled docker binary should be used by system-docker CLI as is implemented in #629. So, I'm going to close this and steal the idea for #629 :wink:

PR #635, opened 2015-11-06
Adding default username and password in readme

chuyd:
Adding default username and password in readme. This will help the users to log when they boot the downloaded image.
Added this comment since I had troubles finding it through the doccumentation. endend using `grep -r password os` to realize that password is set in scripts/isolinux.cfg.  I asked the password in the IRC #rancher early today, and I realize other user had the same issue yesterday.


ibuildthecloud:
Thanks!

Issue #234, opened 2015-04-24
RancherOS should keep the DHCP system container running


sheng-liang:
So it can renew the lease or request new leases.

pwFoo:
That could fix my problem. Because RancherOS sometimes boot without a ip address (static dhcp lease).

Kelindar:
Might be related: I've booted my bare metal RancherOS server without a ethernet cable, plugged the correct one after boot but DHCP didn't assign an address, had to reboot the entire machine. Wouldn't want this to happen in production.

sheng-liang:
I think we need to fix this bug. I just had a failure in network configuration on bare metal server because DHCP server takes too long to return. We need to keep DHCP service retrying.

sheng-liang:
I was able to manually restart network daemon by running
    sudo system-docker restart network

ibuildthecloud:
Two issues here.  1) we don't keep dhcp running 2) we don't respond to kernel events when the link goes up and down.  We should fix them both for sure

ibuildthecloud:
@imikushin I think we are going to have to create a "wait for networking" container.  The reason we don't keep dhcp running is it was a simple way to block until it the network was configured.  If we launch dhcp in the background we don't know when the NIC is configured.  So we can create a blocking "wait for networking" container

imikushin:
@ibuildthecloud dhcpcd is, in fact, "forked to background" when 
`dhcpcd -A4 -e force_hostname=true eth0` command has finished. 

Only when the main process (`netconf`) in the "network" container exits, the container dies and takes dhcpcd with it.

I'm looking how to make "network" a real service solving the issues you outlined.

imikushin:
This issue should be fully resolved (both parts mentioned by @ibuildthecloud) when #639 is merged with rancher/netconf#1 on top of that.

PR #639, opened 2015-11-06
Leave dhcpcd (and netconf) running in the background, add wait-for-network

imikushin:
Leave dhcpcd (and netconf) running in the background, add wait-for-network

See #234

### DIFF
--- a/cmd/network/network.go
+++ b/cmd/network/network.go
@@ -3,6 +3,7 @@ package network
 import (
 	"fmt"
 	"os"
+	"os/exec"
 
 	log "github.com/Sirupsen/logrus"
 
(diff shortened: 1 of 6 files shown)


imikushin:
@ibuildthecloud It's a draft implementation, and has one gotcha: if you want to restart network, you should also restart wait-for-network. This constraint will be dropped in the final impl. 
I guess, I'll also add tests.

imikushin:
@ibuildthecloud The limitation I've described in my previous comment is dropped now. `network` can be restarted w/o a problem.

cmd/network/network.go
```diff
@@ -11,12 +15,43 @@ import (
...
+
 func Main() {
 	args := os.Args
 	if len(args) > 1 {
 		fmt.Println("call " + args[0] + " to load network config from cloud-config.yml")
 		return
 	}
+
+	if strings.HasSuffix(args[0], WAIT_FOR_NETWORK) { // we're wait-for-network
+		go handleTerm()
```
ibuildthecloud: Can you just make this a separate Main() method?  Seems like there is not real value to embedded it here and then you don't need to check `args[0]` because `registerCmd()` just does that for you.

imikushin: Done.

imikushin:
@ibuildthecloud Please also take a look at rancher/netconf#1

PR #647, opened 2015-11-13
os-config.yml: fix the wrong command-volumes configuration

datawolf:
when running command `/usr/bin/ros`, it will get the following
fatal message:
	"Failed to find an entry point for /usr/bin/ros".
The same as the command /usr/sbin/rancherctl /sbin/netconf etc.

This patch fix that, and make it the same sequence as the registerCmd
function in main.go file.

### DIFF
--- a/os-config.yml
+++ b/os-config.yml
@@ -122,19 +122,19 @@ rancher:
       read_only: true
       volumes:
       - /usr/bin/docker:/usr/bin/docker.dist:ro
-      - /usr/bin/ros:/sbin/halt:ro
-      - /usr/bin/ros:/sbin/netconf:ro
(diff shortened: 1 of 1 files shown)


ibuildthecloud:
Hmm.  Are you saying that the order of the bind mounts must match registercmd?  Does that fix something?

imikushin:
It's not the order. We've got `/usr/sbin/ros` in main.go and `/usr/bin/ros` in os-config.yml

Running `/usr/bin/ros` (by full path) gives an error, because the full path registered in main.go is different.

imikushin:
@datawolf Thanks!
LGTM

PR #653, opened 2015-11-18
refactor tests a little bit

imikushin:
also, add dhcpcd test


imikushin:
The dhcpcd test (mentioned above) makes sure dhcpcd keeps running in background after networking has initialized (which it has, when ssh is available). See #234.

PR #659, opened 2015-11-24
Migrate to upstream libcompose

imikushin:
Use upstream docker/libcompose master for service management.
Migrate to Go-1.5.2.
Use glide.yaml to specify dependencies.


imikushin:
@ibuildthecloud @deniseschannon This should be good to go now.

ibuildthecloud:
@imikushin Okay, I'll continue the discussion then https://github.com/docker/libcompose/pull/108#issuecomment-159040628.  We really have to support unquoted true, way too many things will suddenly start failing.

ibuildthecloud:
@imikushin Shoot, I stand corrected, docker-compose does not allow an unquoted true.  Ugh.... This is going to cause a bunch of problems, but ok

imikushin:
Added `./scripts/vendor-cleanup.sh` to remove unneeded packages from ./vendor dir. 

Now there's no need to specify `subpackages` in `glide.yaml` (nor use hacked up glide forks). Just add the needed (top level) packages and run:
```bash
glide up -u --quick --skip-gopath --cache`
./scripts/vendor-cleanup.sh
```
./vendor dir now contains only the necessary packages.

PR #663, opened 2015-11-25
Avoid hardcoded reference to /scripts

robbertkl:
Within a custom provisioning script, I found myself manually running `lay-down-os`, when I discovered it depends on the existence of the `/scripts` directory, which I did not have.

Here's my proposal to avoid this by determining the scripts dir as the one containing `lay-down-os`. To prevent a behaviour change in `seed-data`, I had to modify this script as well.

Please let me know if you have any remarks. Thanks!

### DIFF
--- a/scripts/installer/lay-down-os
+++ b/scripts/installer/lay-down-os
@@ -1,7 +1,9 @@
 #!/bin/bash
 set -e -x
 
-. $(dirname $0)/build.conf
+SCRIPTS_DIR=$(dirname ${0})
+
+. "${SCRIPTS_DIR}/build.conf"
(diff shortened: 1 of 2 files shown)


robbertkl:
Any idea why Drone is claiming the build failed? Where's the error / failed test?

imikushin:
@robbertkl The tests hung, sorry. Drone gave up after an 1 hour timeout.

robbertkl:
Does it need to be restarted so it can pass? Any way to force this?

Is this PR ignored until it passes? Or is it fine to leave it as is..

imikushin:
@robbertkl Thanks! 

LGTM

Issue #339, opened 2015-05-28
Applying cloud-config network configuration without reboot


janeczku:
In order to apply network settings without having to reboot, i currently have to deploy a start script via write_files. Something in the line of
```yaml
write_files:
  - path: /opt/rancher/bin/start.sh
    permissions: 0700
    content: |
      #!/bin/bash
      sudo netconf
```
If i wanted to apply DNS servers as well that script would even be longer.
I think this is kind of a hack and should probably be done programmatically when processing the cloud-config. CoreOS already applies network configuration dynamically (it checks for and applies any networkd units it finds in the cloud-config) and it would be nice to have that supported in RancherOS as well.
What do you think?

/cc @ibuildthecloud @imikushin

imikushin:
@janeczku I'm not sure what you mean here. 

Currently (as of v0.3.1-rc3), RancherOS reads cloud-config and configures network and other system settings on boot accordingly. E.g., DNS servers can be set to OpenDNS's nameservers with the following cloud-config :
```yml
#cloud-config
rancher:
  network:
    dns:
      nameservers:
      - 208.67.222.222
      - 208.67.220.220
```

Can you give an example of what you'd like RancherOS to do? Like, e.g., after changing DNS servers with `ros config set network.dns.nameservers "[8.8.8.8, 8.8.4.4]` to automatically apply them?

janeczku:
@imikushin sorry if i wasn't clear enough. I would like RancherOS to apply any network configurations it finds in the cloud-config immediately without the need to reboot the machine.

imikushin:
@janeczku RancherOS v0.3.1-rc3 and later does exactly that, if I'm not mistaken. I mean, all the configs are read from cloud-config and applied on boot, and the booted machine is fully configured without the need for reboot.

janeczku:
@imikushin I am pretty sure the network settings from cloud-config are still applied only on the next boot, not on the first boot. The network container runs before the cloud-init and i don't see anywhere that the network is configured a second time after processing the cloud-config.

deniseschannon:
@janeczku Just to clarify, are you saying that after doing a `rancheros-install`, you don't want to reboot?

janeczku:
@deniseschannon, yes imagine deploying a cluster and then having to reboot each node individually just so that the network settings get applied. IMHO the network settings in the cloud-config should be applied as they are read/processed.

imikushin:
[...]

Issue #451, opened 2015-07-23
Allow DNS nameservers to be picked up from DHCP before using Google's. 


deniseschannon:
Currently, Google's DNS nameservers are hardcoded.

If someone uses DHCP, the nameservers are not picked up.

We should select DNS nameservers in this order

1. Nameservers from Cloud Config
2. Nameservers set in Rancher.yml (if you used `ros config set`)
3. DHCP nameservers (NEW ADDITION)
4. Use the Google DNS nameservers  that are currently hardcoded

cjellick:
test msg

erichorne:
Upvote this.. It would be really great if the OS picked up the DHCP settings for DNS instead of having to hardcode them in yet-another-place.

deniseschannon:
Verified fixed in v0.4.2 by @ibuildthecloud

PR #669, opened 2015-11-30
Configure DNS nameservers with DHCP or cloud-config

imikushin:
Allow network service to automatically configure DNS from DHCP (by default). 
This can be overridden and configured with values from cloud-config.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -368,7 +368,7 @@
 		},
 		{
 			"ImportPath": "github.com/rancher/netconf",
-			"Rev": "b9aeb623c8ef9551d79efec57ee059658c1ac89c"
+			"Rev": "02925e7cf5a0f0bb0aa5360ee260ef7378e5eff8"
 		},
 		{
 			"ImportPath": "github.com/ryanuber/go-glob",
--- a/Godeps/_workspace/src/github.com/rancher/netconf/types.go
+++ b/Godeps/_workspace/src/github.com/rancher/netconf/types.go
@@ -16,6 +16,7 @@ type InterfaceConfig struct {
 }
 
 type DnsConfig struct {
+	Override    bool     `yaml:"override"`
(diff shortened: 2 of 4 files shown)


imikushin:
Should fix #339. 

This is WIP, because it's based on not yet merged PR #659.

imikushin:
Oh, and this needs rancher/netconf#3

deniseschannon:
https://github.com/rancher/os/issues/451

qrpike:
Still having this issue.

dszryan:
using version 0.4.2
i have set nameservers and override via ros (confirmed values are written in clould-config.yml)
but the values in resolv.conf are still dhcp server values, even after reboot or sudo netconf

and unable to resolve domain names
image

image

qrpike:
@dszryan I had the same issues and literally just spend all night trying to get it working. Ended up having to put the cloud-config.yml as a write-file. Inception basically. Example:

https://gist.github.com/qrpike/cfd5245d4459a2d6efd8

Pretty hacky and I don't like it at all, but it's the cleanest thing I could come up with. I've tried about everything lol

dszryan:
i swapped over from docker itself, coz of this issue.... 
and this wasnt an issue v0.4.1, i am rolling back to the older version until its fixed

mtrofimm:
@qrpike do you if it's possible to include dynamic some dynamic param into var/lib/rancher/conf/cloud-config.yml?
something like: $(wget -O- http://169.254.169.254/current/meta-data/local-ipv4 into eth1 address?

djmaze:
Having the same problems. I don't know how this could even work with DHCP. The DHCP client (`dhcpcd`) is running in the background all the time, overwriting any values set manually in resolv.conf. 

As far as I can see, the client's configuration (`dhcpcd.config`) has to be adjusted for this to work as well. Using `nooption domain_name_servers, domain_name, domain_search` or similar.

PR #684, opened 2015-12-15
Use drone-0.4 for CI and migrate build to dapper

imikushin:
Use drone-0.4 for CI and migrate build to dapper


ibuildthecloud:
@imikushin I like this.  Can you rebase?

PR #701, opened 2015-12-22
Use a subdirectory of the state partition

ibuildthecloud:
The allows you to do 
```yaml
#cloud-config
rancher:
  state:
     directory: ros
```
And the sub directory `ros` will be used as the root of RancherOS in the partition `RANCHER_STATE`


ibuildthecloud:
@deniseschannon This needs docs

PR #702, opened 2015-12-22
 Farewell PID one

ibuildthecloud:
This change no longer runs Docker as PID 1.  Instead PID 1 is a very
simple zombie reaper and Docker is moved as a child of that PID.


deniseschannon:
This change sounds like it requires docs changes. 

http://docs.rancher.com/os/#how-this-works

ibuildthecloud:
@deniseschannon Yes, a lot of docs need to change.

PR #735, opened 2016-01-18
Multi-arch build: amd64, arm

imikushin:
Build for amd64, arm

ARM build: `dapper make ARCH=arm rootfs` produces ./dist/artifacts/rootfs.tgz which should be extracted into the **root** partition (formatted as ext4) of a Raspberry Pi SD card (just be sure to remove everything from that root partition before). 

Make sure you add `init=/init rw`to the kernel boot params in cmdline.txt on Raspberry Pi SD card **boot** partition (formatted as FAT).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,16 +1,22 @@
-include build.conf
 FORCE_PULL := 0
 DEV_BUILD  := 0
+ARCH       := amd64
+
+include build.conf
+include build.conf.$(ARCH)
 
 
-installer: minimal
-	docker build -t $(IMAGE_NAME):$(VERSION) .
-
-bin/rancheros:
+bin/ros:
 	mkdir -p $(dir $@)
-	go build -tags netgo -installsuffix netgo -ldflags "-X github.com/rancher/os/config.VERSION=$(VERSION) -linkmode external -extldflags -static" -o $@
-	strip --strip-all $@
+	ARCH=$(ARCH) VERSION=$(VERSION) ./scripts/mk-ros.sh $@
 
+build/host_ros: bin/ros
+	mkdir -p $(dir $@)
+ifeq "$(ARCH)" "amd64"
+	ln -sf ../bin/ros $@
+else
+	ARCH=amd64 VERSION=$(VERSION) ./scripts/mk-ros.sh $@
+endif
 
 pwd := $(shell pwd)
 include scripts/build-common
@@ -22,6 +28,11 @@ assets/docker:
 	chmod +x $@
 
 
+ifdef COMPILED_KERNEL_URL
+
+installer: minimal
+	docker build -t $(IMAGE_NAME):$(VERSION) .
+
 $(DIST)/artifacts/vmlinuz: $(BUILD)/kernel/
 	mkdir -p $(dir $@)
 	mv $(BUILD)/kernel/boot/vmlinuz* $@
@@ -29,20 +40,40 @@ $(DIST)/artifacts/vmlinuz: $(BUILD)/kernel/
 
 $(BUILD)/kernel/:
 	mkdir -p $@
-	([ -e "$(COMPILED_KERNEL_URL)" ] && cat "$(COMPILED_KERNEL_URL)" || curl -L "$(COMPILED_KERNEL_URL)") | tar -xzf - -C $@
+	curl -L "$(COMPILED_KERNEL_URL)" | tar -xzf - -C $@
 
 
-$(BUILD)/images.tar: bin/rancheros
-	FORCE_PULL=$(FORCE_PULL) ./scripts/mk-images-tar.sh
-
-
-$(DIST)/artifacts/initrd: bin/rancheros assets/docker $(BUILD)/kernel/ $(BUILD)/images.tar
+$(DIST)/artifacts/initrd: bin/ros assets/docker $(BUILD)/kernel/ $(BUILD)/images.tar
 	mkdir -p $(dir $@)
-	DFS_IMAGE=$(DFS_IMAGE) DEV_BUILD=$(DEV_BUILD) ./scripts/mk-initrd.sh
+	ARCH=$(ARCH) DFS_IMAGE=$(DFS_IMAGE) DEV_BUILD=$(DEV_BUILD) ./scripts/mk-initrd.sh $@
 
 
 $(DIST)/artifacts/rancheros.iso: minimal
 	./scripts/mk-rancheros-iso.sh
+
+all: minimal installer iso
+
+initrd: $(DIST)/artifacts/initrd
+
+minimal: initrd $(DIST)/artifacts/vmlinuz
+
+iso: $(DIST)/artifacts/rancheros.iso $(DIST)/artifacts/iso-checksums.txt
+
+test: minimal
+	cd tests/integration && tox
+
+.PHONY: all minimal initrd iso installer test
+
+endif
+
+
+$(BUILD)/images.tar: build/host_ros
+	ARCH=$(ARCH) FORCE_PULL=$(FORCE_PULL) ./scripts/mk-images-tar.sh
+
+
+$(DIST)/artifacts/rootfs.tar.gz: bin/ros assets/docker $(BUILD)/images.tar
+	mkdir -p $(dir $@)
+	ARCH=$(ARCH) DFS_IMAGE=$(DFS_IMAGE) DEV_BUILD=$(DEV_BUILD) IS_ROOTFS=1 ./scripts/mk-initrd.sh $@
 
 
 $(DIST)/artifacts/iso-checksums.txt: $(DIST)/artifacts/rancheros.iso
@@ -52,13 +83,6 @@ $(DIST)/artifacts/iso-checksums.txt: $(DIST)/artifacts/rancheros.iso
 version:
 	@echo $(VERSION)
 
-all: minimal installer iso
+rootfs: $(DIST)/artifacts/rootfs.tar.gz
 
-minimal: $(DIST)/artifacts/initrd $(DIST)/artifacts/vmlinuz
-
-iso: $(DIST)/artifacts/rancheros.iso $(DIST)/artifacts/iso-checksums.txt
-
-test: minimal
-	cd tests/integration && tox
-
-.PHONY: build-all minimal iso installer version bin/rancheros integration-tests
+.PHONY: rootfs version bin/ros
--- a/build.conf.arm
+++ b/build.conf.arm
@@ -1,0 +1,1 @@
+DOCKER_BINARY_URL=https://github.com/rancher/docker/releases/download/v1.10.0-ros_arm/docker-1.10.0
--- a/init/bootstrap.go
+++ b/init/bootstrap.go
@@ -1,7 +1,6 @@
 package init
 
 import (
-	"os"
 	"syscall"
 
 	"fmt"
@@ -61,7 +60,7 @@ func stopDocker(c chan interface{}) error {
 	c <- struct{}{}
 	<-c
 
-	return os.RemoveAll(config.DOCKER_SYSTEM_HOME)
+	return nil
 }
 
 func bootstrap(cfg *config.CloudConfig) error {
--- a/init/init.go
+++ b/init/init.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+	"syscall"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/rancher/docker-from-scratch"
@@ -17,6 +18,9 @@ import (
 
 const (
 	STATE string = "/state"
+
+	TMPFS_MAGIC int64 = 0x01021994
+	RAMFS_MAGIC int64 = 0x858458f6
 )
 
 var (
@@ -140,15 +144,17 @@ func tryMountState(cfg *config.CloudConfig) error {
 }
 
 func tryMountAndBootstrap(cfg *config.CloudConfig) (*config.CloudConfig, error) {
-	if err := tryMountState(cfg); !cfg.Rancher.State.Required && err != nil {
-		return cfg, nil
-	} else if err != nil {
-		return cfg, err
-	}
+	if isInitrd() {
+		if err := tryMountState(cfg); !cfg.Rancher.State.Required && err != nil {
+			return cfg, nil
+		} else if err != nil {
+			return cfg, err
+		}
 
-	log.Debugf("Switching to new root at %s %s", STATE, cfg.Rancher.State.Directory)
-	if err := switchRoot(STATE, cfg.Rancher.State.Directory, cfg.Rancher.RmUsr); err != nil {
-		return cfg, err
+		log.Debugf("Switching to new root at %s %s", STATE, cfg.Rancher.State.Directory)
+		if err := switchRoot(STATE, cfg.Rancher.State.Directory, cfg.Rancher.RmUsr); err != nil {
+			return cfg, err
+		}
 	}
 	return mountOem(cfg)
 }
@@ -170,10 +176,21 @@ func getLaunchConfig(cfg *config.CloudConfig, dockerCfg *config.DockerConfig) (*
 	return &launchConfig, args
 }
 
+func isInitrd() bool {
+	var stat syscall.Statfs_t
+	syscall.Statfs("/", &stat)
+	return int64(stat.Type) == TMPFS_MAGIC || int64(stat.Type) == RAMFS_MAGIC
+}
+
 func RunInit() error {
 	os.Setenv("PATH", "/sbin:/usr/sbin:/usr/bin")
-	// Magic setting to tell Docker to do switch_root and not pivot_root
-	os.Setenv("DOCKER_RAMDISK", "true")
+	if isInitrd() {
+		log.Debug("Booting off an in-memory filesystem")
+		// Magic setting to tell Docker to do switch_root and not pivot_root
+		os.Setenv("DOCKER_RAMDISK", "true")
+	} else {
+		log.Debug("Booting off a persistent filesystem")
+	}
 
 	initFuncs := []config.CfgFunc{
 		func(c *config.CloudConfig) (*config.CloudConfig, error) {
--- a/os-config_arm.yml
+++ b/os-config_arm.yml
@@ -2,7 +2,7 @@ hostname: rancher
 rancher:
   bootstrap:
     state-script:
-      image: rancher/os-statescript:v0.4.3-dev
+      image: rancher/os-statescript:v0.4.3-dev_arm
       labels:
         io.rancher.os.detach: "false"
         io.rancher.os.scope: system
@@ -18,7 +18,7 @@ rancher:
       - /usr/bin/ros:/usr/bin/ros:ro
       - /usr/share/ros:/usr/share/ros:ro
     udev-bootstrap:
-      image: rancher/os-udev:v0.4.3-dev
+      image: rancher/os-udev:v0.4.3-dev_arm
       environment:
       - BOOTSTRAP=true
       labels:
@@ -34,7 +34,7 @@ rancher:
       - /lib/firmware:/lib/firmware
   autoformat:
     autoformat:
-      image: rancher/os-autoformat:v0.4.3-dev
+      image: rancher/os-autoformat:v0.4.3-dev_arm
       labels:
         io.rancher.os.detach: "false"
         io.rancher.os.scope: system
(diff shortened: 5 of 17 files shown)


os-config_arm.yml
```diff
@@ -0,0 +1,327 @@
...
+      - all-volumes
+      volumes:
+      - /sys/fs/cgroup:/host/sys/fs/cgroup
+  system_docker:
+    args: [daemon, --log-opt, max-size=25m, --log-opt, max-file=2, -s, overlay, -b, docker-sys,
+      --fixed-cidr, 172.18.42.1/16, --restart=false, -g, /var/lib/system-docker, -G, root,
+      -H, 'unix:///var/run/system-docker.sock', --userland-proxy=false]
+  upgrade:
+    url: https://releases.rancher.com/os/releases.yml
+    image: rancher/os
```
luxas: `rancher/os:arm`?

scripts/mk-ros.sh
```diff
@@ -0,0 +1,19 @@
...
+
+strip_bin=$(which strip)
+if [ "${ARCH}" == "arm" ]; then
+  export GOARM=6
+  export CC=/usr/bin/arm-linux-gnueabihf-gcc
+  export CGO_ENABLED=1
+  strip_bin=/usr/arm-linux-gnueabihf/bin/strip
+fi
+GOARCH=${ARCH} go build -tags netgo -installsuffix netgo -ldflags "-X github.com/rancher/os/config.VERSION=${VERSION} -linkmode external -extldflags -static" -o ${ros}
+${strip_bin} --strip-all ${ros}
```
luxas: Why are you doing this? I suppose for saving disk space...

Just asking cause I've read that it could harm `go` binaries and I don't know if it's supported by the go team.

It could be hard to debug if something goes wrong...

Makefile
```diff
@@ -1,27 +1,29 @@
...
 	mkdir -p $(dir $@)
-	go build -tags netgo -installsuffix netgo -ldflags "-X github.com/rancher/os/config.VERSION=$(VERSION) -linkmode external -extldflags -static" -o $@
-	strip --strip-all $@
+	ARCH=$(ARCH) VERSION=$(VERSION) ./scripts/mk-ros.sh $@
 
+./ros: bin/ros
+ifeq "$(ARCH)" "amd64"
+	ln -sf bin/ros ./ros
+else
+	ARCH=amd64 VERSION=$(VERSION) ./scripts/mk-ros.sh $@
```
luxas: I don't understand what this code does...
Maybe provide some comments?

luxas:
I'm on this train. Will help you with what I can.
BTW I'm porting kubernetes to ARM, ARM64 and ppc64le: kubernetes/kubernetes#17981

os-config_arm.yml
imikushin: Right. But it should be rather releases_arm.yml

Makefile
imikushin: So, the build works in a regular docker container (on a amd64 linux server) and we need ./ros command for a later build step - extracting system container images from os-config.yml (or os-config_arm.yml). We can use the same binary when building for amd64, but need to build a separate one when building for arm.

imikushin:
@luxas Thanks for your feedback! Can you please take another look at the code: I've added building a rootfs.tgz archive (see above for usage).

Makefile
luxas: Ah, OK. But could you make a comment then (so folks understand `./ros` is for the release process, not a result)

build.conf.arm
```diff
@@ -0,0 +1 @@
+COMPILED_KERNEL_URL=https://github.com/rancher/os-kernel/releases/download/HypriotOS-4.1.12-v7+/linux-4.1.12-hypriotos-v7.-arm.tar.gz
```
luxas: Has this kernel all necessary libs?

os-config_arm.yml
```diff
@@ -0,0 +1,328 @@
...
+      - all-volumes
+      volumes:
+      - /sys/fs/cgroup:/host/sys/fs/cgroup
+  system_docker:
+    args: [daemon, --log-opt, max-size=25m, --log-opt, max-file=2, -s, overlay, -b, docker-sys,
+      --fixed-cidr, 172.18.42.1/16, --restart=false, -g, /var/lib/system-docker, -G, root,
+      -H, 'unix:///var/run/system-docker.sock', --userland-proxy=false]
+  upgrade:
+    url: https://releases.rancher.com/os/releases_arm.yml
+    image: rancher/os
```
luxas: I also meant that the name of the image might change to `rancher/os-arm` or something. Not so important right now but anyway

scripts/mk-initrd.sh
```diff
@@ -8,24 +19,36 @@ INITRD_DIR=${BUILD}/initrd
...
-                                                   usr
+docker export ${DFS_ARCH} | tar xvf - -C ${INITRD_DIR} --exclude=usr/bin/dockerlaunch \
+                                                       --exclude=usr/share/git-core   \
+                                                       --exclude=usr/bin/git          \
+                                                       --exclude=usr/bin/ssh          \
+                                                       --exclude=usr/libexec/git-core \
+                                                       usr
+
+if [ "$IS_ROOTFS" == "1" ]; then
+  DFS=$(docker run -d --privileged -v /lib/modules/$(uname -r):/lib/modules/$(uname -r) ${DFS_IMAGE})
```
luxas: Doesn't this just run _on_an `arm` host?

imikushin: The build runs on a "normal" amd64 host :smile:  That's why we can create arm (and potentially other architectures) containers, but not run them during the build.

build.conf.arm
imikushin: This build actually runs on RPi, so I guess, yes it does. What libs do you mean?

I'm going to try Raspbian kernel, just for funzies, as well. Hypriot was the easiest to start with as docker is known to work on it.

luxas: Yes, just asked.
Then, in the future, I think it's best to make an own kernel that has exactly everything RancherOS needs (and doesn't need), but I agree it's best to start with hypriot

DieterReuter: @imikushin I think the standard Raspbian kernel is still missing some important Docker tuning settings. But feel free to ask. It's not too complicated to customise and compile your own kernel for the Raspberry Pi. We've just open sourced all of these bits in our repo https://github.com/hypriot/rpi-kernel and you could build your own kernel pretty easily as well, just modify the .config and build it. :smirk:

scripts/mk-initrd.sh
luxas: Yes, I understand that we're running this on an amd64 host, but missed that `${DFS_IMAGE}` was `rancher/docker` (thought it was `rancher/docker-arm` and was therefore concerned)

FYI, it's possible to run and build ARM docker images on `amd64' hosts too: https://resin.io/blog/building-arm-containers-on-any-x86-machine-even-dockerhub/

Makefile
ibuildthecloud: I find this also confusing and especially don't like that we are creating a file in `./`.  Can you create a file like `./build/host-ros`.  Calling it host aligns a bit with cross compilation setups where the host and target are different.

Makefile
```diff
@@ -1,27 +1,29 @@
...
+endif
 
 pwd := $(shell pwd)
 include scripts/build-common
 
 
-assets/docker:
```
ibuildthecloud: We shouldn't remove this.  I want rancher/os to directly reference Docker.  It's a PITA to have to build a seperate `rancher/docker` image just to have a different Docker binary.

init/one.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 func pidOne() error {
-	c := make(chan os.Signal)
+	c := make(chan os.Signal, 10)
```
ibuildthecloud: Not needed anymore, this was fixed on master (and set to 2048)

os-config_arm.yml
ibuildthecloud: Yeah, we honestly don't know how to upgrade arm yet...

os-config_arm.yml
```diff
@@ -0,0 +1,328 @@
+hostname: rancher
```
ibuildthecloud: I'm going to lose sleep over the fact that this file is 90% the same as `os-config.yml`.   Do we have the technology to not duplicate so much?

scripts/mk-ros.sh
ibuildthecloud: We've been doing this for awhile and no problems yet :)  It doesn't remove the line info for go panics, so we haven't had any problem debugging.  But yes, this is for file size.

ibuildthecloud:
Okay, so two major-ish changes.

1) Don't pull docker from rancher/docker, but instead have build.conf reference a docker to download, just like before.
2) We shouldn't reference the HypriotOS kernel.  For ARM all I want to release right now is the rootfs.tar that does not have any kernel/firmware.  This will be a completely generic ARM userspace.  We later need to build a post step that combines rootfs.tar + board specific image + board specific kernel/firmware to produce an SD card image.  Honestly I'd like to see if we can join forces with @DieterReuter on this one and hopefully share some work.

init/init.go
```diff
@@ -168,10 +171,21 @@ func getLaunchConfig(cfg *config.CloudConfig, dockerCfg *config.DockerConfig) (*
 	return &launchConfig, args
 }
 
+func isInitrd() bool {
+	var stat syscall.Statfs_t
+	syscall.Statfs("/", &stat)
+	return stat.Type == 0x01021994
```
ibuildthecloud: This should be checking for both tmps and ramfs.  There is probably a constant somewhere for this, but if there isn't one can you declare a constant and avoid the magic number.  Just makes it obscure what you're doing here.

init/init.go
```diff
@@ -168,10 +171,21 @@ func getLaunchConfig(cfg *config.CloudConfig, dockerCfg *config.DockerConfig) (*
...
+}
+
 func RunInit() error {
 	os.Setenv("PATH", "/sbin:/usr/sbin:/usr/bin")
-	// Magic setting to tell Docker to do switch_root and not pivot_root
-	os.Setenv("DOCKER_RAMDISK", "true")
+	if isInitrd() {
+		log.Info("Booting off an in-memory FS")
```
ibuildthecloud: Can you remove this log message or move it to debug.  I don't want to see this on every boot.  And change `FS` to `file system`.

init/init.go
```diff
@@ -168,10 +171,21 @@ func getLaunchConfig(cfg *config.CloudConfig, dockerCfg *config.DockerConfig) (*
...
+		// Magic setting to tell Docker to do switch_root and not pivot_root
+		os.Setenv("DOCKER_RAMDISK", "true")
+	} else {
+		log.Info("Booting off a persistent FS")
```
ibuildthecloud: Same comment as the above log message.

init/init.go
imikushin: Agree on making it a named constant. This value is the TMPFS_MAGIC
from here: http://man7.org/linux/man-pages/man2/statfs.2.html

I tried to check for RAMFS_MAGIC too, but it's out of range for int32 which is the datatype for stat.Type on arm (it's int64 on amd64). So I had to remove that check for now.

ibuildthecloud: Those constants are 32bit, how can it be out of range?

imikushin: try building it and you'll get:
```
+ export GOARM=6
+ GOARM=6
+ export CC=/usr/bin/arm-linux-gnueabihf-gcc
+ CC=/usr/bin/arm-linux-gnueabihf-gcc
+ export CGO_ENABLED=1
+ CGO_ENABLED=1
+ strip_bin=/usr/arm-linux-gnueabihf/bin/strip
+ GOARCH=arm
+ go build -tags netgo -installsuffix netgo -ldflags '-X github.com/rancher/os/config.VERSION=v0.4.3-dev -linkmode external -extldflags -static' -o bin/ros
# github.com/rancher/os/init
init/init.go:177: constant 2240043254 overflows int32
Makefile:13: recipe for target 'bin/ros' failed
make: *** [bin/ros] Error 2
```

Makefile
imikushin: I'm moving `./ros` to `build/host_ros`. To enable this I'm also adding the default entrypoint in main.go.

imikushin:
rootfs **without** modules/firmware is unbootable: 
```
ERRO[0002] Failed to create bridge docker-sys: operation not supported
```
And then it's 
```
INFO[xxxx] Waiting for Docker at unix:///var/run/system-docker.sock
```
every second, forever.

imikushin:
Okay, not forever: it panics in ~5 minutes, probably because of the timeout waiting for system-docker to become available.

ibuildthecloud:
@imikushin of course it's not bootable without the modules/firmware.  They have to be added to the SD image

init/init.go
ibuildthecloud: Well shoot, not sure what is the right incarnation here.  You need to set the constant to a uint32 I just don't know what happens when you cast it.

ibuildthecloud: Does just unit32(0xF00) fix it?

ibuildthecloud: Oops meant int32(0xF00)

imikushin:
@ibuildthecloud I've separated rootfs.tgz and usrlib.tgz (which contains modules and firmware) in the last commit. So now you can run `dapper make ARCH=arm rootfs usrlib` and get whatever you want :smile:

ibuildthecloud:
@imikushin I'd rather not produce the usrlib.  Even with the usrlib the user still needs to take a series of steps to produce a bootable sd image.  So I do see much point to it and the usrlib is tied to a rpi.  I'd much rather for this release not refer to the hypriot kernel at all and just produce the rootfs.  As a convince for this release we can manually assemble a full sd image.  But the next release we need to automate that as a separate step but I want that to be more generic so we can target more boards.

djmaze:
Just got this working on an XU4 by using the rootfs alongside an existing bootloader, /boot and /lib/modules directories. Thanks for the work so far! :+1:

os-config_arm.yml
luxas: Fair enough, we'll come to that later...

init/init.go
luxas: I don't know for sure what to do here, but some conversation is probably necessary. 
In #559 it was fixed by first casting the constant `syscall.LINUX_REBOOT_CMD_POWER_OFF` to `uint` and then back to `int`.

Makefile
```diff
@@ -32,13 +41,22 @@ $(BUILD)/kernel/:
...
+	ARCH=$(ARCH) FORCE_PULL=$(FORCE_PULL) ./scripts/mk-images-tar.sh
+
+
+$(DIST)/artifacts/initrd: bin/ros assets/docker $(BUILD)/kernel/ $(BUILD)/images.tar
+	mkdir -p $(dir $@)
+	ARCH=$(ARCH) DFS_IMAGE=$(DFS_IMAGE) DEV_BUILD=$(DEV_BUILD) ./scripts/mk-initrd.sh $@
 
 
-$(DIST)/artifacts/initrd: bin/rancheros assets/docker $(BUILD)/kernel/ $(BUILD)/images.tar
+$(DIST)/artifacts/rootfs.tgz: bin/ros assets/docker $(BUILD)/kernel/ $(BUILD)/images.tar
```
luxas: Why not `.tar.gz`? Just asking... isn't it more common?

luxas:
Awesome @djmaze! I'll try this too ASAP on a RPi 1 and 2.

I agree with @ibuildthecloud that the first RancherOS on ARM release should feature only a `rootfs.tar.gz` and `alpha` testers have to add e.g. the Hypriot kernel, `/boot` and `/lib/modules` directories.

Then we could discuss support for other types of boards and maybe `arm64` :smile:

os-config_arm.yml
luxas: Maybe tag images `rancher/os-xxx-ARCH:VERSION` and replace `ARCH` and `VERSION` via a `sed` command before using the file? That would solve the problem...

luxas: I think that's the best way, because think when we'll add support for `arm64`, it gets even worse

ibuildthecloud: @luxas there is also a difference in that acpid is not included in ARM.  but again, nothing some sed can't fix.  I personally have never liked that the version has to be updated in this file.  One point you bring up though is should the arch be in the image name or label.  Right now it's in the label.  I think we went with that approach because it seemed more inline with say a dpkg versioning scheme where the arch is not in the package name.

Makefile
imikushin: NP, `.tar.gz` it is

init/bootstrap.go
```diff
@@ -61,7 +60,7 @@ func stopDocker(c chan interface{}) error {
 	c <- struct{}{}
 	<-c
 
-	return os.RemoveAll(config.DOCKER_SYSTEM_HOME)
+	return nil
```
ibuildthecloud: This change concerns me a bit, have to think through it a bit more.

ibuildthecloud:
@imikushin I'll merge on one condition.  You absolutely promise that we will not maintain a os-config.yml and os-config_arm.yml going forward.  So next PR will consolidate these?

init/init.go
imikushin: This should be fixed now. Both tmpfs and ramfs are being checked.

imikushin:
@ibuildthecloud Sure! I'll consolidate these in a next PR.

ibuildthecloud:
@imikushin ok

scripts/mk-initrd.sh
```diff
@@ -8,29 +19,48 @@ INITRD_DIR=${BUILD}/initrd
...
 
-cp -rf ${BUILD}/kernel/lib ${INITRD_DIR}/usr
+if [ "$IS_ROOTFS" == "0" ]; then
+  cp -rf ${BUILD}/kernel/lib ${INITRD_DIR}/usr/
+fi
 cp assets/docker           ${INITRD_DIR}/usr/bin/docker
-cp ${BUILD}/images.tar     ${INITRD_DIR}/usr/share/ros
-cp os-config.yml           ${INITRD_DIR}/usr/share/ros/
-cp bin/rancheros           ${INITRD_DIR}/usr/bin/ros
+if [ "$IS_ROOTFS" == "0" ]; then
```
luxas: Why wouldn't images be included in the `rootfs`?

scripts/mk-initrd.sh
```diff
@@ -8,29 +19,48 @@ INITRD_DIR=${BUILD}/initrd
...
+                                                       --exclude=usr/bin/ssh          \
+                                                       --exclude=usr/libexec/git-core \
+                                                       usr
+
+if [ "$IS_ROOTFS" == "1" ]; then
+  DFS=$(docker run -d --privileged -v /lib/modules/$(uname -r):/lib/modules/$(uname -r) ${DFS_IMAGE})
+  trap "docker rm -fv ${DFS_ARCH} ${DFS}" EXIT
+  docker exec -i ${DFS} docker load < ${BUILD}/images.tar
+  docker stop ${DFS}
+  docker run --rm --volumes-from=${DFS} debian:jessie tar -c -C /var/lib/docker ./image | tar -x -C ${INITRD_DIR}/var/lib/system-docker
```
luxas: never mind, the images are loaded here in the rootfs... maybe some comment that describes what it does?

luxas:
OK, got it partially working on a Pi 2.
It has some rough edges though: ~1GB size and boots in ~60s
However, it's very annoying to have it working but not be able to log in.
None of `rancher/rancher`, `root/root`, `pi/raspberry`, `root/hypriot` work.

@djmaze How did you log in?

@imikushin @ibuildthecloud ??

DieterReuter:
@luxas users and passwords should be defined in the rootfs and not in the kernel package, so there must be the Rancher defaults apply - nothing from Hypriot.

luxas:
@DieterReuter Yeah, I know. But I had to test it anyway :)
And yeah, I used `/boot/` and `/lib/modules` from latest Hypriot image

djmaze:
@luxas You need to add a `cloud-config.yml` to `/var/lib/rancher/conf/` in the rootfs before booting the first time. That file should include the `ssh_authorized_keys` [as described in the manual](http://docs.rancher.com/os/cloud-config/).

djmaze:
Btw, the [image builders by hypriot](https://github.com/search?utf8=%E2%9C%93&q=user%3Ahypriot+image-builder&type=Repositories&ref=searchresults) help a lot in that. I used one adjusted for my XU4, replaced the rootfs tgz, adjusted the boot.ini and added the cloud config before building the final image.

luxas:
Thanks @djmaze! Was only thinking about using my keyboard to the Pi, so didn't think about ssh at all. But now when I've added my computer to ssh, it works!! :smile:
Will try to run kubernetes on it now

PR #748, opened 2016-02-04
do not print success info when loading images error

datawolf:
when the client.LoadImage return an error, there should not be print
success info like the following:

    "INFO[0000] Done loading images from /usr/share/ros/images.tar"

This patch fix it by move the log.Infof() after the err check.


imikushin:
LGTM

PR #754, opened 2016-02-07
Fix bug in install script

oillio:
lsblk -o label will list the labels of the block devices. Using name only returns /dev/sda, etc.
In some cases, labels may be repeated (eg - RAID configurations).  It is safer to use greater than 0.


ibuildthecloud:
Thanks!

PR #772, opened 2016-02-23
SELinux tools command

joshwget:


cmd/control/selinux.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+func selinuxCommand() cli.Command {
+	app := cli.Command{}
+	app.Name = "selinux"
+	app.Usage = "Launch SELinux tools container."
+	app.Action = func(c *cli.Context) {
+		argv := []string{"system-docker", "run", "-it", "--privileged", "--rm", "-v",
+			"/etc/selinux:/etc/selinux", "-v", "/var/lib/selinux:/var/lib/selinux",
+			"-v", "/usr/share/selinux:/usr/share/selinux", "rancher/selinuxtools", "bash"}
```
ibuildthecloud: Can you change the image to `"rancher/selinuxtools:"+config.Version`?  Can you also add all the same volume mounts that are in the console, basically what is in `all-volumes` in `os-config.yml`?

PR #795, opened 2016-03-01
ARM installer

imikushin:
To build a RancherOS installer image for ARM architecture you need to set COMPILED_KERNEL_URL in `build.conf.arm`, e.g.:
```
COMPILED_KERNEL_URL=https://github.com/imikushin/os-kernel/releases/download/Estuary-4.1.6-ros_arm64/linux-4.1.6-ros_arm64.tar.gz
```

Then run `./build-arm-installer.sh`


.arm.docker-env.sample
```diff
@@ -0,0 +1,2 @@
+DOCKER_HOST="tcp://192.168.2.2:2375"
```
luxas: Could you add a comment what this file does? It took a while before I understood the purpose, that it redirects docker commands on your machine to a Raspberry Pi

Dockerfile.arm
```diff
@@ -0,0 +1,13 @@
+FROM resin/armel-debian:jessie
```
luxas: Why not `resin/rpi-raspbian:jessie`?

Dockerfile.arm
```diff
@@ -0,0 +1,13 @@
...
+ENV DEBIAN_FRONTEND noninteractive
+ENV ARCH arm
+
+RUN apt-get update && apt-get install -y parted kexec-tools
+
+COPY ./scripts/installer /scripts
```
luxas: Should be `/scripts/`

Dockerfile.arm
ibuildthecloud: I'd prefer we use `armel/debian` as then seems to be the more "standard" image.  @luxas What is the primary difference with the raspbian image?  This is just for the installer image, the one that lays out the partitions and copies the kernel to install from ISO or upgrade.  Basically both of these flows don't currently work on RPi.  Currently we are working on a [Estuary D02](http://open-estuary.org/d02-2/) which is a very PC-like ARM board that support UEFI/grub so it works a lot like an intel server.  So I'm not sure what the raspbian image would provide for us, we'd prefer this to be generic ARM as possible.

.arm.docker-env.sample
ibuildthecloud: Agreed.  I don't think we have instructions anywhere on how to actually build arm.

build-arm-installer.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+set -e
+
+cd $(dirname $0)
+
+[ -f "./.arm.docker-env" ] || ( echo "cp ./.arm.docker-env.sample ./.arm.docker-env, edit to your settings (use an ARM docker host), run this again"; exit 1 )
```
ibuildthecloud: 3 lines of script are more readable than 1 runon :)

build-arm-installer.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
ibuildthecloud: Why is this script in root, where as all other scripts are in `./scripts`?

ibuildthecloud: Is this not kicked off my make?

Dockerfile.arm
luxas: `GOARM=5` == `armel`, `GOARM=7`== `armhf`, `GOARCH=arm64` == `aarch64/arm64`
Raspberry Pi 1: `GOARM=6`, **`armhf`** (Other `GOARM=6` boards doesn't have `armhf` support)
Raspberry Pi 2: `GOARM=7`, `armhf`
Raspberry Pi 3, Estuary: `GOARCH=arm64`
`GOARCH=arm64` may run `GOARM=7` binaries, `GOARM=7` may run `GOARM=6` and `GOARM=5` binaries, but not the other way around.
The thing is that often `GOARM=7` and `armhf` is considered equal, but it's not.

All `armhf/*` images on Docker Hub are therefore `ARMv7` and all `armel/*` are `ARMv5`. RPi 1 is between those,  and `resin/rpi-raspbian` is `GOARM=6` and `armhf`

OK, that was hard to try to explain. For maximum support though, go with official `armel/debian`, as there is no significant perfomance boost in having `armhf/debian` and dropping support for some boards.

Makefile
```diff
@@ -36,31 +33,31 @@ ifdef COMPILED_KERNEL_URL
 installer: minimal
 	docker build -t $(IMAGE_NAME):$(VERSION) .
 
-$(DIST)/artifacts/vmlinuz: $(BUILD)/kernel/
+dist/artifacts/vmlinuz: build/kernel/
```
luxas: Good to ditch `$(DIST)` btw. `dist/`, `build/` and the like is much easier to understand

Dockerfile.arm
imikushin: For now we will not build an installer for arm, only for amd64 and arm64. The one for arm64 is based on aarch64/debian:jessie and it comes with a limitation: it does not install grub.

PR #808, opened 2016-03-16
print proper tips when running `ros` or `system-docker` without using sudo 

datawolf:
`ros` and `system-docker` command requires user to be the root user,
when we running them without using sudo, it will print the follow:

```
[rancher@rancher ~]$ ros os list
ERRO[0000] Failed to read /var/lib/rancher/conf/cloud-config.d: open
/var/lib/rancher/conf/cloud-config.d: permission denied
ERRO[0000] Error reading config files                    err=open
/var/lib/rancher/conf/cloud-config.yml: permission denied
files=[/var/lib/rancher/conf/cloud-config.yml]
ERRO[0000] Failed [1/4] 25%
ERRO[0000] Failed to load config
......
......
FATA[0000] open /var/lib/rancher/conf/cloud-config.yml: permission
denied
```

and

```
[rancher@rancher ~]$ system-docker restart docker
Failed to kill container(docker): Cannot connect to the Docker daemon.
Is the docker daemon running on this host?
```
this patch make the tips more clear and simple.

### DIFF
--- a/cmd/control/cli.go
+++ b/cmd/control/cli.go
@@ -3,6 +3,7 @@ package control
 import (
 	"os"
 
+	log "github.com/Sirupsen/logrus"
 	"github.com/codegangsta/cli"
 	"github.com/rancher/os/config"
 )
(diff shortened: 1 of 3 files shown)


imikushin:
LGTM
@ibuildthecloud what do you think?

PR #810, opened 2016-03-16
Support SELinux on arm/arm64

imikushin:
Support SELinux on arm/arm64


imikushin:
WIP until rancher/docker-from-scratch#50 is merged and vendored

PR #816, opened 2016-03-22
fix gen-os-config.sh error

datawolf:
Need to source build.conf.${ARCH} file when we generate os-config.yml
after the pr(https://github.com/rancher/os/pull/815) merged.

this patch fix it.


imikushin:
@datawolf Thanks! LGTM

PR #821, opened 2016-03-22
Replace curl with wget in Makefile

imikushin:
Replace curl with wget in Makefile

Because curl exits cleanly on HTTP errors (like 404 Not Found) and output filled with the error page content

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,12 +23,12 @@ endif
 
 assets/docker:
 	mkdir -p $(dir $@)
-	curl -L "$(DOCKER_BINARY_URL)" > $@
+	wget -O - "$(DOCKER_BINARY_URL)" > $@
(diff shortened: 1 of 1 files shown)


vincent99:
That's not a very nice thing to do to Josh...

imikushin:
@vincent99 Josh has already done exactly that! @joshwget right?

ibuildthecloud:
How about you just add `-f` to `curl`

joshwget:
Unfortunately, I can only do as I am programmed :100:

imikushin:
@ibuildthecloud From curl manual regarding `-f, --fail`:
> This method is not fail-safe and there are occasions where non-successful response codes will slip through, especially when authentication is involved (response codes 401 and 407)

Issue #672, opened 2015-12-01
ubuntu-console would not run on bare metal server


jackjacek:
I'm experiencing an issue with latest rancher (v0.4.1) across two servers. I tried enabling ubuntu-console using cloud-config.yml at installation time, but it always fails on boot: Failed to load ubuntu-console : Get https://raw.githubusercontent.com/rancherio/os-services/v0.4.1/u/ubuntu-console.yml: dial tcp: lookup raw.githubusercontent.com on 8.8.4.4:53: read udp 8.8.4.4:53: network unreachable. I tried editing /var/lib/rancher/conf/cloud-config.d/user_config.yml and only left the necessary ssh keys section, after rebooting I had to enable ubuntu-console from command line (not sure why it did not persist) & after rebooting still getting the same error. I can verify pulling rancher/os-ubuntuconsole image worked without an issue.

far-blue:
I'm getting this exact same issue using VirtualBox with a vm setup through docker-machine. Sometimes it works but mostly it doesn't.

liyimeng:
I have exactly the same issue, but it turns out that my cooperation network has blocked the access to google DNS server. After redirect the DNS query traffic, all works fine.

deniseschannon:
We have tested in other versions of RancherOS and this should be fixed now.

PR #825, opened 2016-03-25
Run tests on qemu-system-aarch64

imikushin:


datawolf:
Without these patches, the ARM64 CI can not run successfully. hope to merge these to master branch.

imikushin:
@datawolf @ibuildthecloud Please take a look. This should be ready to go.

compose/project.go
```diff
@@ -177,7 +177,7 @@ func newCoreServiceProject(cfg *config.CloudConfig, network bool) (*project.Proj
 
 	go func() {
 		for event := range projectEvents {
-			if event.EventType == project.EventContainerStarted && event.ServiceName == "network" {
+			if event.EventType == project.EventContainerStarted && event.ServiceName == "ntp" {
```
ibuildthecloud: @imikushin why ntp?  why not "wait-for-network"?

os-config.tpl.yml
```diff
@@ -105,7 +105,7 @@ rancher:
         io.rancher.os.detach: "false"
         io.rancher.os.reloadconfig: "true"
         io.rancher.os.scope: system
-        io.rancher.os.after: cloud-init-pre, wait-for-network
+        io.rancher.os.after: cloud-init-pre, wait-for-network, ntp
```
ibuildthecloud: Why do you depend on ntpd? The reason I ask is because if you need the system time to be set, there is no real guarantee it will be because we launch ntp in the background.

scripts/shell.sh
```diff
@@ -4,4 +4,4 @@ set -e
 cd $(dirname $0)/..
 . ./scripts/dapper-common
 
-exec dapper -s
+exec dapper -d -s
```
ibuildthecloud: Do you want to disable debug?

compose/project.go
imikushin: Because HTTPS certs appear "not yet valid"

os-config.tpl.yml
imikushin: There's no guarantee, of course, but a pretty good chance. Otherwise ubuntu-console or other external services that are loaded via HTTPS just can't load on boot. 

We could make wait-for-network depend on ntp if that looks better.

Another approach could be to ditch HTTPS and use hash-sums (cached in the config), though this would make configuration a bit bloated and prone to errors. But that's probably a topic for another discussion.

scripts/shell.sh
imikushin: No, I don't. But if you think it's better without it, okay.

deniseschannon:
https://github.com/rancher/os/issues/672

PR #827, opened 2016-03-26
Fix lay-down-os shell error

datawolf:
the current code for lay-down-os will cause the following error when
install rancheros to disk.

```
-bash: [: missing `]'
```


imikushin:
LGTM

Issue #771, opened 2016-02-19
ros os upgrade does not count on local image


liyimeng:
I would like to test upgrade. So I build an image and which tagged as rancher/os:v0.4.4-dev. I load this image into system-docker as below
<pre>
[rancher@rancher ~]$ sudo system-docker images
REPOSITORY               TAG                 IMAGE ID            CREATED             SIZE
rancher/os               v0.4.4-dev          df84018a6caa        30 minutes ago      193.6 MB
rancher/os               v0.4.3              eed7c8ab50fd        3 days ago          193.1 MB
rancher/os-preload       v0.4.3              983c005fe53f        3 days ago          25.65 MB
rancher/os-console       v0.4.3              d9b2845438df        3 days ago          25.66 MB
rancher/os-udev          v0.4.3              8dc9eee7501f        2 weeks ago         25.65 MB
rancher/os-syslog        v0.4.3              987960440665        2 weeks ago         25.65 MB
rancher/os-statescript   v0.4.3              24355446800e        2 weeks ago         25.65 MB
rancher/os-state         v0.4.3              0fe21afc3049        2 weeks ago         25.65 MB
rancher/os-ntp           v0.4.3              3e2d57d4ae21        2 weeks ago         25.65 MB
rancher/os-network       v0.4.3              5288f2eb944e        2 weeks ago         25.65 MB
rancher/os-docker        v0.4.3              6a4e2f959df2        2 weeks ago         25.65 MB
rancher/os-cloudinit     v0.4.3              5c47e775e016        2 weeks ago         25.65 MB
rancher/os-autoformat    v0.4.3              00182a66713c        2 weeks ago         25.65 MB
rancher/os-acpid         v0.4.3              0fa901101944        2 weeks ago         25.65 MB
</pre>
I then run upgrade:
<pre>
[rancher@rancher ~]$ sudo ros os upgrade -i rancher/os:v0.4.4-dev
INFO[0000] Project [once]: Starting project             
INFO[0000] [0/1] [os-upgrade]: Starting                 
INFO[0000] Rebuilding os-upgrade                        
INFO[0000] [1/1] [os-upgrade]: Started                  
INFO[0000] Project [once]: Project started              
Pulling repository docker.io/rancher/os
ERRO[0008] Failed to pull image rancher/os:v0.4.4-dev: Tag v0.4.4-dev not found in repository docker.io/rancher/os 
FATA[0008] Tag v0.4.4-dev not found in repository docker.io/rancher/os
</pre>

[...]

PR #832, opened 2016-03-28
Only pull upgrade image if not found locally

joshwget:
This solves two problems relating to #771.

1. If a local image exists and there is a remote image with the same name, `ros os upgrade` will pull and overwrite the local image.
2. If a local image exists and there isn't a remote image with the same name, `ros os upgrade` will give an error saying that the upgrade image does not exist.


liyimeng:
When this one is to be merged?

PR #833, opened 2016-03-28
Use images for the current architecture (vs just amd64)

### LINKED ISSUE #828: [os] sudo ros install -d /dev/vda can not install rancheros successfully on ARM64/ARM
On ARM/ARM64, when we type the command `sudo ros install -d /dev/vda`, it will install the x86_64 RancherOS not the ARM/ARM64 RancherOS.

so please fix it.

imikushin:
Install from the image for the current architecture (not for x86_64 only)

Fix #828

### DIFF
--- a/cmd/control/install.go
+++ b/cmd/control/install.go
@@ -63,7 +63,7 @@ func installAction(c *cli.Context) {
 		log.WithFields(log.Fields{"err": err}).Fatal("ros install: failed to load config")
 	}
 	if image == "" {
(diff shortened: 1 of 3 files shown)


imikushin:
Added fix to the similar issue #835

ibuildthecloud:
@imikushin Do these need to be set as compiler flags/constants?

imikushin:
@ibuildthecloud No, they don't.

PR #839, opened 2016-03-29
Dockerfile.dapper: make sure to update apt repos before fetching build deps and sources

imikushin:
Dockerfile.dapper: make sure to update apt repos before fetching build deps and sources.

Also, use dapper -d flag in scripts/shell.sh for better visibility

### DIFF
--- a/scripts/shell.sh
+++ b/scripts/shell.sh
@@ -4,4 +4,4 @@ set -e
 cd $(dirname $0)/..
 . ./scripts/dapper-common
 
-exec dapper -s
+exec dapper -d -s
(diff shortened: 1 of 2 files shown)


scripts/shell.sh
```diff
@@ -4,4 +4,4 @@ set -e
 cd $(dirname $0)/..
 . ./scripts/dapper-common
 
-exec dapper -s
+exec dapper -d -s
```
ibuildthecloud: probably don't want to enable debug mode.

imikushin: Dapper is not too verbose with debug. And you probably want to know the arguments to docker if you're using ./scripts/shell.sh

I've added this to the PR because I found myself adding this temporarily a few times already.

PR #842, opened 2016-03-30
Include VirtualBox guest modules

joshwget:


ibuildthecloud:
Can you change the vbox tar to have the kernel version in the path such that we don't need the extra environment variable.

PR #845, opened 2016-03-31
fix the suffix error when the target is rootfs

datawolf:


datawolf:
I use the ARM64 machine(Estuary board) to build the RancherOS.

imikushin:
LGTM

PR #852, opened 2016-04-01
Test build for all architectures

joshwget:


ibuildthecloud:
Does this mean the build is broken??

joshwget:
No, I think the build is currently okay. I was trying to add a check that all architectures build to CI, but I'm not really sure what's going on. I'm hoping to finish this with @imikushin tomorrow.

ibuildthecloud:
@joshwget I agree this would be great.  We can also hook up ARM64 CI.

Issue #846, opened 2016-03-31
cloud-config not being applied even after reboot


linxit:
I'm using terraform to provision RancherOS 0.4.3 in AWS, and trying to get the host to register with our Rancher server automatically. I'm passing various configs via cloud-config which aren't being set, even after a reboot.

The config goes into /var/lib/rancher/conf/cloud-config.d/boot.yml, exactly as provided. But very little of it is being applied - just hostname, ssh_authorized_keys and proxy settings are working.

The DNS and route settings are just ignored. If I then copy these into /var/lib/rancher/conf/cloud-config.yml (or run ros config set ...) and reboot, then it starts working.

I've seen https://github.com/rancher/os/issues/339, but at the moment I can't even get it working after a reboot without some manual tweaking.

```
#cloud-config
hostname: <my_hostname>
ssh_authorized_keys:
- ssh-rsa 
  <pubkey>
rancher:
  network:
    dns:
      search:
        - <domain1>
        - <domain2>
      nameservers:
        - <ns>
    interfaces:
      eth0:
        dhcp: true
        gateway: <gateway>
system_docker:                    
    environment:                    
    - HTTP_PROXY=<proxy:port>  
    - HTTPS_PROXY=<proxy:port>   
    - NO_PROXY=<.domain>
  docker:                           
    environment:                 
    - HTTP_PROXY=<proxy:port>
    - HTTPS_PROXY=<proxy:port>
    - NO_PROXY=<.domain>
  services:
    rancher-agent1:
      image: rancher/agent:v0.11.0
      command: sudo docker run -d --privileged -v /var/run/docker.sock:/var/run/docker.sock -v /var/lib/rancher:/var/lib/rancher rancher/agent:v0.11.0 <rancher_url> 
  services_include:
    ubuntu-console: true
    debian-console: false
```

deniseschannon:
This is due to the fact that we can't download the user-data until networking is setup, but we don't reload the networking (i.e. dns/routes) with any changes from the user-data cloud-config.

linxit:
I've been trying to work around this by adding a custom /etc/dhcpcd.conf with the user-data, but although this is loaded it's ignored by the system even after a reboot. 

So my custom dhcpcd.conf has static settings for routers, domain_name_servers and domain_search (with the relevant option settings from the default removed).

I didn't expect it to "just work" due to the current open bug, but surely it should be read and applied on boot?

deniseschannon:
Verified this is working in v0.4.4-rc3.

[...]

Issue #758, opened 2016-02-11
/usr/sbin/netconf command hangs forever


janeczku:
> $ netconf -h

```
call netconf to load network config from cloud-config.yml
```

> $ sudo netconf

```
(...)
killall: wait-for-network: no process killed
```

Command hangs forever and has to be force closed with Control+C

vovimayhem:
Mine too :(

vovimayhem:
OK @janeczku are you trying to restart the network so the configuration gets applied? Have you tried `sudo ros service restart` instead??

janeczku:
@vovimayhem Yep. That's what i ended up using, too! :grinning: Nevertheless i would really like to know what's the "official" way to restart the network.

ibuildthecloud:
'sudo ros service restart network` is the "official" way to restart networking.  In 0.4.3 this should work better but we don't clean up things really.  For example if you had IP 1.1.1.1 and configured 2.2.2.2 and restarted networking; we would add 2.2.2.2 and leave 1.1.1.1 on there.

vovimayhem:
@ibuildthecloud I have the following question: Do `sudo ros config set rancher.network.dns.nameserver [208.67.222.222,208.67.220.220] && sudo ros service restart network` is expected to update the `/etc/resolv.conf` file accordingly?

I expected it to, but it didn't...

janeczku:
@vovimayhem There is a pending issue concerning that https://github.com/rancher/os/issues/720

deniseschannon:
Verified with v0.4.4-rc4.

`sudo netconf` doesn't hang.

`sudo ros config set rancher.network.dns.nameserver [208.67.222.222,208.67.220.220] && sudo ros service restart network` updates the networking.

PR #862, opened 2016-04-06
Apply network configurations again after cloud-init

joshwget:
Depends on rancher/netconf#10 and rancher/os-images#73

Adds a flag `-daemon` to netconf. If true, netconf will behave as it does now and run dhcpd as a daemon. If false, netconf will exit after applying network configurations.

Adds two new system services to apply network configurations again after cloud-init. The two new services are called network and wait-for-network, and the previous network and wait-for-network services are now network-pre and wait-for-network-pre. The new services are essentially the same as the old ones, except they occur after cloud-init.

New order for services:
1. cloud-init-pre
2. network-pre
3. wait-for-network-pre
4. ntp
5. cloud-init
6. network
7. wait-for-network

#846 
#758 
#339

### DIFF
--- a/cmd/network/network.go
+++ b/cmd/network/network.go
@@ -2,7 +2,7 @@ package network
 
 import (
 	"bufio"
-	"fmt"
+	"flag"
 	"io/ioutil"
 	"os"
 	"os/exec"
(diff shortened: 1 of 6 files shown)


janeczku:
Another issue to track: https://github.com/rancher/os/issues/867

joshwget:
@janeczku I didn't actually know this PR fixed any hostname issues. Have you noticed any of those problems resolved?

janeczku:
@joshwget The way i understood the report in #867 is that there is not an issue with the hostname at all. It's just that DHCP is run before the hostname is set to the cloud-config value. That's why he sees all his hosts reported with "rancher" hostname in the DHCP server. This would also be resolved by this PR since reloading the network would reload DHCP as well.

PR #864, opened 2016-04-07
fix lay-down-os error

datawolf:
option `-k` does not need an argument and `-a` need one.


datawolf:
Test error was not caused by this patch

imikushin:
LGTM
Thanks @datawolf !

PR #865, opened 2016-04-07
Warn before pull image when upgrading to the currently running version

datawolf:
If user get the Warn and give up the upgrade. we should not pull the image.
so move the Warn code before `container.Pull()` method.


joshwget:
I think moving the confirmation before the image pull makes sense. If we're already going to move one confirmation though, I think we should also move the other one.

Another issue with this is that the user should not have to confirm anything if they use the `--stage` flag.

datawolf:
@joshwget The current patch will have nothing to confirm if user use the `--stage` flag.

joshwget:
This should have been fixed in #955. Thanks @datawolf!

PR #870, opened 2016-04-08
Check the validity of the service name

datawolf:
if user want to switch to `ubuntu-console` and type the wrong service name `ubuntu-consile`, the command `sudo ros service enable ubuntu-consile` run with no error. after the reboot, the console is still the busybox console.

It is better to Warn user that the `ubuntu-consile` is no a valid service name.

This patch also check the validity of the service name when disable and delete service.


joshwget:
I think this is a great change. Thanks for the PR @datawolf!

PR #876, opened 2016-04-12
Revert "Check the validity of the service name"

joshwget:
Reverts rancher/os#870

The idea behind this PR is good, but the current implementation breaks some of the functionality with [custom services](http://docs.rancher.com/os/configuration/system-services/#adding-custom-system-services). We'll try to implement this with a different approach for v0.4.5.


datawolf:
@joshwget ,on version v0.4.5, it seems that rancheros does not check the validity of the service name.

Issue #855, opened 2016-04-04
os-console does not upgrade when we upgrade the RancherOS


datawolf:
When I install RancherOS v0.4.2 and switch to ubuntu-console with the following command:

```
$ sudo ros os install -d /dev/sda
$ sudo reboot
$ sudo ros service enable ubuntu-console
$ sudo reboot
```
Now, I enter the ubuntu console. And then I upgrade the RancherOS to `v0.4.3` using the command `sudo ros os upgrade`. After the RancherOS reboot, all system service are upgrade to the `v0.4.3` version, but the version of ubuntu-console is still the `v0.4.2`.

Is this right? I think RancherOS should pull the new version of the console when start after the reboot in this situation.

liyimeng:
To my understanding, ubuntu-console is persistent console, it should be keep as it is during upgrade.

imikushin:
@datawolf If we upgrade the console container, all its state would be lost. Because you cannot just update the image of a container: you rebuild the container entirely.

datawolf:
@imikushin I think that whether to keep the console container should be decided by user. we can add a parameter for command `ros os upgrade` to control whether to upgrade the console container(the default is not) during RancherOS upgrade.

datawolf:
@imikushin how about to add a parameter for command `ros os upgrade`

deniseschannon:
Verified with v0.4.5-rc1 to v0.4.5-rc2

PR #890, opened 2016-04-18
Add upgrade flag to upgrade persistent console

joshwget:
Adds the flag `upgrade-console` to `ros os upgrade`. This will set the new config flag `force_console_rebuild`, which will force a console rebuild on next boot. After rebuilding the console, the flag is set back to false.

### DIFF
--- a/cmd/control/os.go
+++ b/cmd/control/os.go
@@ -58,6 +58,10 @@ func osSubcommands() []cli.Command {
 					Name:  "append",
 					Usage: "kernel args to append by kexec",
 				},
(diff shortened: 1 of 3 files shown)


liyimeng:
Do we migrate data from old container if a force rebuild to happen?

joshwget:
No, this should only be used if you're willing to part ways with your data :smile:

liyimeng:
@joshwget OK, might be useful in a dev environment.

deniseschannon:
https://github.com/rancher/os/issues/855

PR #897, opened 2016-04-23
Remove `-D` when we set rancher.debug=false

datawolf:
when we set rancher.debug to true, the current code also set `-D` flag
to system-docker and user-docker daemon.
and then we change rancher.debug to false, the `-D` flag is also exist
by using command `sudo ros config get rancher.docker`.

this patch fix it.

### DIFF
--- a/config/disk.go
+++ b/config/disk.go
@@ -61,6 +61,13 @@ func LoadConfig() (*CloudConfig, error) {
 		if !util.Contains(cfg.Rancher.SystemDocker.Args, "-D") {
 			cfg.Rancher.SystemDocker.Args = append(cfg.Rancher.SystemDocker.Args, "-D")
 		}
+	} else {
+		if util.Contains(cfg.Rancher.Docker.Args, "-D") {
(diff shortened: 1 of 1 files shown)


config/disk.go
```diff
@@ -61,6 +61,13 @@ func LoadConfig() (*CloudConfig, error) {
 		if !util.Contains(cfg.Rancher.SystemDocker.Args, "-D") {
 			cfg.Rancher.SystemDocker.Args = append(cfg.Rancher.SystemDocker.Args, "-D")
 		}
+	} else {
+		if util.Contains(cfg.Rancher.Docker.Args, "-D") {
+			cfg.Rancher.Docker.Args = util.FilterStrings(cfg.Rancher.Docker.Args, func(x string) bool { return x != "-D" })
+		}
+		if !util.Contains(cfg.Rancher.SystemDocker.Args, "-D") {
```
imikushin: @datawolf You probably want to skip `!` here if you want to test if `cfg.Rancher.SystemDocker.Args` contains `-D`

imikushin:
Just one line comment above. Otherwise looks good.

datawolf:
@imikushin Sorry for the mistake, the patch have been updated.

imikushin:
LGTM

PR #904, opened 2016-04-30
Move DHCP from network into its own service

joshwget:
Depends on rancher/netconf#14 and rancher/os-images#80

Moves DHCP functionality out of network-pre and network services. This allows them to run with `io.rancher.os.detach: "false"`, which eliminates the need for wait-for-network-pre and wait-for-network.


ibuildthecloud:
Cool, let's discuss this on Monday.

Issue #899, opened 2016-04-25
user console not receiving HTTP_PROXY settings


deniseschannon:

deniseschannon:
This is working for @cloudnautique with v0.4.5-rc2.

PR #917, opened 2016-05-06
HTTP proxy settings can be set in cloud config

joshwget:
Depends on rancher/netconf#16


deniseschannon:
https://github.com/rancher/os/issues/899

Issue #616, opened 2015-10-23
Do not drop default values in "ros config export"


sheng-liang:
"ros config export" should print default values, otherwise people won't know what the configuration really is.

For example, if I set "eth*: dhcp: false", ros config export will print "eth*: {}"

ibuildthecloud:
@imikushin When reading the files for export we shouldn't marshal to a CloudConfig type because it looses the ability to print default values.  We need to unmarshall to a map (merge them) and then print the map.

faddat:
+1

imikushin:
Currently, `ros config export --full` prints full configuration, including values not explicitly set by the user (_aka_ default). 

@sheng-liang @ibuildthecloud Should we make this behaviour the default 
(i.e. _without_ specifying `--full`)?

faddat:
Thank you!

(I think it probably should be default, otherwise what is the purpose of
exporting the config?)

deniseschannon:
Verified with custom build

`sudo ros config set docker.tls false`

In the `sudo ros config export`, the default value is still shown as I had set it explicitly.

PR #922, opened 2016-05-08
Simplify configuration

joshwget:
#616

### DIFF
--- a/config/disk.go
+++ b/config/disk.go
@@ -15,14 +15,9 @@ import (
 	"github.com/rancher/os/util"
 )
 
-var osConfig *CloudConfig
-
-func NewConfig() *CloudConfig {
-	if osConfig == nil {
-		osConfig, _ = ReadConfig(nil, true, OsConfigFile, OemConfigFile)
-	}
-	newCfg := *osConfig
-	return &newCfg
+func NewConfig() map[interface{}]interface{} {
+	osConfig, _ := readConfig(nil, true, OsConfigFile, OemConfigFile)
+	return osConfig
 }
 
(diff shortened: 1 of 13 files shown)


config/disk.go
```diff
@@ -85,36 +88,63 @@ func CloudConfigDirFiles() []string {
...
+	if cfg.Rancher.Debug {
+		log.SetLevel(log.DebugLevel)
+		if !util.Contains(cfg.Rancher.Docker.Args, "-D") {
+			cfg.Rancher.Docker.Args = append(cfg.Rancher.Docker.Args, "-D")
+		}
+		if !util.Contains(cfg.Rancher.SystemDocker.Args, "-D") {
+			cfg.Rancher.SystemDocker.Args = append(cfg.Rancher.SystemDocker.Args, "-D")
+		}
+	} else {
+		if util.Contains(cfg.Rancher.Docker.Args, "-D") {
```
ibuildthecloud: This should be removed.  This was added because of a side effect.  If rancher.debug is false we shouldn't strip out `-D`.

config/types.go
```diff
@@ -1,10 +1,11 @@
 package config
 
 import (
+	"runtime"
```
ibuildthecloud: Why runtime import?

joshwget: It was already there, just got moved around a little. Looks like like it's used here https://github.com/joshwget/os/blob/bb25737ade67ae25cd4391460aadec179df80bbd/config/types.go#L64

ibuildthecloud: ah I see.

Issue #741, opened 2016-01-26
Allow ability to mount additional devices


deniseschannon:

zhaohb:
@deniseschannon  now I am studying rancheros，I can make iso myself, but I want to know how rancheros manage system service ,and What work has been done for this,can you help me?  thank you very much

pwFoo:
Hello @zhaohb,

that should help:
http://docs.rancher.com/os/configuration/system-services/

I created a first convoy plugin system service for testing:
https://github.com/rancher/os/issues/706

zhaohb:
@pwFoo thank you for you letter, http://docs.rancher.com/os/configuration/system-services/ I had already seen, but what I want to know is that the deeper level of knowledge, for example system-docker is how to manage and start udev and NTP services, and the relationship between a variety of services is how to solve

zhaohb:
@pwFoo for example a services rely on the Lib Library in the B service，how to solve ,thank you very much

zhaohb:
@deniseschannon  @pwFoo 
Can you give me some advice？ thank you very much!

joshwget:
Confirmed with rancher/os:v0.5.0-rc2

PR #923, opened 2016-05-08
Additional mounts can be set in cloud config

joshwget:
#741


ibuildthecloud:
Switch the to standard `mount` field.  Also you also support swap.

Issue #933, opened 2016-05-12
Allow ability to support Amazon's hostname (aa.bb.cc.dd.ee) for TLS


onnoll:
**RancherOS Version: 0.4.4**

**Where are you running RancherOS? AWS**

I tried to configure TLS following this [instructions](http://docs.rancher.com/os/configuration/setting-up-docker-tls/) but it didn't work. I receive this error when I try to connect from my local machine.

> An error occurred trying to connect: Get https://10.0.1.230:2376/v1.23/version: x509: certificate is valid for 127.0.0.1, not 10.0.1.230

 even when I made `sudo ros tls gen -H 10.0.1.230` to generate the client certs.

curx:
same for RancherOS v0.4.5, seems the ros tls generate --server|-s produce an TLS /etc/docker/tls/server-cert.pem with wrong X509v3 Subject Alternative Name, no matter what ros config options are selected.

e.g.  see http://docs.rancher.com/os/rancheros-tools/ros/tls/

[rancher@rancher ~]$  sudo ros tls generate -s --hostname 172.0.0.1 --hostname localhost --hostname foo.example.com

[rancher@rancher ~]$ sudo openssl x509 -noout -text -in /etc/docker/tls/server-cert.pem 
```
Certificate:
    Data:
        Version: 3 (0x2)

...(truncated)...

        X509v3 extensions:
            X509v3 Key Usage: critical
                Digital Signature, Key Encipherment, Key Agreement
            X509v3 Extended Key Usage: 
                TLS Web Client Authentication, TLS Web Server Authentication
            X509v3 Basic Constraints: critical
                CA:FALSE
            X509v3 Subject Alternative Name: 
                DNS:*, DNS:*.*, DNS:*.*.*, DNS:*.*.*.*, IP Address:127.0.0.1
```

deniseschannon:
Verified with v0.5.0-rc1 using an internal AMI.

PR #954, opened 2016-05-24
Enable more dots in RancherOS docker host names (for Amazon EC2)

imikushin:
Enable more dots in RancherOS docker host names (for Amazon EC2)


imikushin:
Fixes #933

imikushin:
Wildcards are convenient, but we want the users to be explicit.

Closing.

Issue #883, opened 2016-04-15
Update order of operations in upgrade 


deniseschannon:
In v0.4.4, we download the latest image, then ask if you want ot upgrade to a specific version. The problem is that you don't know what the latest release version is and it doesn't ask for confirmation until after it's downloaded an image.

The order should be:

1. Ask for confirmation
2. Download the image
3. Continue upgrade
4. Ask for rebot

deniseschannon:
We should fix this because it will automatically download the image before asking if you want to upgrade to the same version.

deniseschannon:
Verified with custom build. 

We don't ask again after we pull the image whether or not you still want to upgrade, but we ask before downloading the image, which was what I primarily wanted fixed.

PR #955, opened 2016-05-24
ros os upgrade: ask for confirmation first, then download and upgrade

imikushin:
ros os upgrade: ask for confirmation first, then download the image and run upgrade

#883

### DIFF
--- a/cmd/control/os.go
+++ b/cmd/control/os.go
@@ -216,6 +216,16 @@ func startUpgradeContainer(image string, stage, force, reboot, kexec bool, upgra
 		}
 	}
 
+	fmt.Printf("Upgrading to %s\n", image)
+	confirmation := "Continue"
+	imageSplit := strings.Split(image, ":")
+	if len(imageSplit) > 1 && imageSplit[1] == config.VERSION+config.SUFFIX {
+		confirmation = fmt.Sprintf("Already at version %s. Continue anyway", imageSplit[1])
+	}
+	if !force && !yes(in, confirmation) {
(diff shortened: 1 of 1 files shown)


joshwget:
We should probably only ask for confirmation once in the case of upgrading to the current version. This will ask for confirmation twice. When using the `--stage` flag it'll prompt with the "Upgrading to ..." message, which is a little odd. Maybe we don't need to prompt for confirmation when staging.

imikushin:
@joshwget Updated.

cmd/control/os.go
```diff
@@ -216,6 +216,16 @@ func startUpgradeContainer(image string, stage, force, reboot, kexec bool, upgra
 		}
 	}
 
+	fmt.Printf("Using %s\n", image)
```
joshwget: "Using" sort of reads as if that's the version that's currently running. I think "Upgrading to" was a bit more clear.

imikushin: Changed

cmd/control/os.go
```diff
@@ -216,6 +216,16 @@ func startUpgradeContainer(image string, stage, force, reboot, kexec bool, upgra
 		}
 	}
 
+	fmt.Printf("Upgrading to %s\n", image)
+	confirmation := "Continue"
+	imageSplit := strings.Split(image, ":")
```
ibuildthecloud: In general we should start using the distribution code to parse image names because there's a tons of gotchas.  Right now this is fine though.

PR #956, opened 2016-05-24
Preserve `/usr/local` across reboots

imikushin:
Preserve `/usr/local` across reboots.

Makes it easier to customize docker (by dropping custom docker binaries into `/usr/local/bin`) without forcing to switch to a persistent console.

### DIFF
--- a/os-config.tpl.yml
+++ b/os-config.tpl.yml
@@ -336,6 +336,7 @@ rancher:
       volumes:
       - /home:/home
       - /opt:/opt
+      - /usr/local:/usr/local
     docker:
       image: {{.OS_REPO}}/os-docker:{{.VERSION}}{{.SUFFIX}}
       labels:


joshwget:
If we do this we can update [this test](https://github.com/rancher/os/blob/master/tests/integration/rostest/test_05_custom_docker_in_persistent_console.py) to not use a persistent console. It'll make it run a little faster.

ibuildthecloud:
I'm not really for this.  `/opt` wqs originally designed for this.  We should probably put a custom docker in `/opt/bin/`

imikushin:
/usr/local is normally used to hold stuff, well, local to this machine. So
users might expect it to be persistent. And /usr/local/bin is normally on
$PATH

ibuildthecloud:
I'm going to close.  We should have users put custom dockers in `/opt`.  Currently `/usr/local/bin` is not in the PATH btw.

Issue #937, opened 2016-05-13
v0.4.5: cannot switch to ubuntu console


sheng-liang:
Installing 0.4.5 on bare metal. After installing the OS, switch to Ubuntu console, and then reboot. After reboot it fell back to busybox after seeing error messages on the console. Console screen shots attached.

![image](https://cloud.githubusercontent.com/assets/9485278/15263495/7187024e-191f-11e6-97b0-e0a67395be51.png)

![image](https://cloud.githubusercontent.com/assets/9485278/15263512/92839e58-191f-11e6-9126-f2b63659f872.png)

sheng-liang:
Why are we trying to fetch something from the network? Ubuntu console container has already been pulled. We should not be attempting any network connection during machine start (and especially not before the network has been configured as is the case here.)

sheng-liang:
I tried the same steps on a VMware VM and it worked. So the issue is only reproducible on one of our bare metal servers with 10Gbe NIC.

joshwget:
Tested with a v0.5 RC and confirmed that this is fixed!

PR #958, opened 2016-05-27
Replace wait-for-network services with retries

joshwget:
The biggest change here is that `network`/`wait-for-network` will no longer guarantee an Internet connection. Instead, all services that require network continue to retry or eventually timeout.

Internally the Go net library will cache DNS settings and only refresh them every five seconds. This can now slow down the boot process, and so a patch is included to add a manual refresh.

`network-pre` no longer exits after getting an IP for one interface. Instead, it's run as a daemon similar to `network` which allows dhcpcd to configure remaining interfaces. This should resolve issues such as #937. `network-pre` is stopped once `network` is reached.


Dockerfile.dapper
```diff
@@ -101,9 +101,11 @@ RUN if [ -n "${!VBOX_MODULES_URL}" ]; then \
     ;fi
 
 # Install Go
+COPY assets/go-dnsclient.patch ${DAPPER_SOURCE}
```
ibuildthecloud: Don't copy to ${DAPPER_SOURCE} use /usr/src for this.

ibuildthecloud: Or `${DOWNLOADS}`

tests/integration/rostest/test_10_network_from_url.py
```diff
@@ -28,9 +28,6 @@ def test_network_interfaces_conf(qemu):
 ip link show dev br0
 ip link show dev br0.100 | grep br0.100@br0
 ip link show dev eth1.100 | grep 'master br0'
-ip link show dev eth6 | grep 'master bond0'
```
ibuildthecloud: What has this been removed?

PR #975, opened 2016-06-02
Disable docker server TLS cert auto-generation

imikushin:
Disable docker server TLS cert auto-generation

Because users should be explicit about their docker TLS certs.

Also, write the key and cert files when `ros tls gen` is run


ibuildthecloud:
@imikushin Can you make passing certs in the cloud-config on boot work also.  Also this PR needs a rebase.

imikushin:
Updated. Keys and certs from cloud-config will get written to files in `/etc/docker/tls` on docker system service start if rancher.docker.tls is true.

joshwget:
Can you add an integration test with `rancher.docker.tls` true in the cloud config?

imikushin:
@joshwget You were right: I added the test and found a bug! Now it's good to go.

PR #977, opened 2016-06-02
Bump ubuntu-console to 16.04, add man

imikushin:
Bump ubuntu-console to 16.04, add man

Also, minor image cleanups (use cached layers to build faster).


ibuildthecloud:
Why man?

Issue #507, opened 2015-08-17
GCE cloud-config metadata user-data script execution


fentas:
According to this http://docs.rancher.com/os/cloud-config/
if I start a gce instance with `--metadata-from-file user-data=${__dirname}/bootstrap/startup` it should execute the content of startup script "[...] If the file starts with #!<interpreter> (e.g., #!/bin/sh) [...]".

I don't know maybe I misunderstand, but it should execute this if the system boots up or `cloud-init` is somewhere triggered?
If so (on boot up), I have the feeling it's not implemented yet or something is not working.

If I execute the script manually on the booted system everything works like a charm.
Maybe I miss something? Thanks!

*I found this in `system-docker.log` relavant?*
```log
time="2015-08-17T18:41:30.278849189Z" level=info msg="GET /containers/cloud-init/json" 
time="2015-08-17T18:41:30.278945290Z" level=error msg="Handler for GET /containers/{name:.*}/json returned error: no such id: cloud-init" 
time="2015-08-17T18:41:30.278963826Z" level=error msg="HTTP Error" err="no such id: cloud-init" statusCode=404 
```
*also `/var/log/messages`*
```log
Aug 17 18:58:20 rancher docker/79d874d96daf[1]: time="2015-08-17T18:58:20Z" level=error msg="Unrecognized cloud-init\n" 
Aug 17 18:58:20 rancher docker/79d874d96daf[1]: time="2015-08-17T18:58:20Z" level=info msg="Merging /var/lib/rancher/conf/cloud-config.d/user_config.yml" 
Aug 17 18:58:20 rancher docker/79d874d96daf[1]: time="2015-08-17T18:58:20Z" level=info msg="Writing to /var/lib/rancher/conf/cloud-config-processed.yml" 
Aug 17 18:58:22 rancher docker/61c3f45d3f11[1]: time="2015-08-17T18:58:22Z" level=info msg="Merging cloud-config from meta-data and user-data" 
```

ps. Somehow I have to execute a script on startup. Maybe other solution? I saw that there is a ticket on gce agent support.

imikushin:
Hi, so the message `Unrecognized cloud-init` (from `/var/log/messages`) is relevant. The code that produces this message should also have dumped your user-data, right after `/n`, so this looks pretty odd.

To make sure user data is actually being fetched, can you look for messages containing `Fetching user-data from datasource ` and `Fetching meta-data from datasource `? 

[...]

PR #978, opened 2016-06-03
Use GCE metadata service

imikushin:
To fetch cloud-config and other metadata (project and instance specific ssh keys)


imikushin:
#507

PR #990, opened 2016-06-09
LUKS/cryptsetup, BTRFS, LVM2, mdadm, iSCSI, NFS for real use cases

tetatetit:
Also:
- Switched 4.4 kernel which has essentially newer BTRFS version as result more thorough and reliable
- Added EXTRA_MODULES_URL build var allowing to build with extra kernel modules needed to boot on some bare metal (real case with my server, that's why added :-) )
- Allowing BTRFS for docker unsticking it from overlay driver by default (prevented it to start from mounted BTRFS as /var/lib/docker without changing config manually)
-  Added ssh_authorized_keys='["ssh_rsa fookey1== foo@bar.com", "ssh_rsa fookey2== foo@bar.com]' kernel param support mapped to ros/cloud-init scheme appropritely, as well as all other with spaces (single- or double-quoted like shell args), except write_file but it can be enabled too if makes sence.
- Temporary switched to repo path to my forked till my os-base LUKS, BTRFS and other storage related chages are not merged to original.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,12 @@ else
 	touch $@
 endif
 
+assets/extra.tar.gz:
+ifdef EXTRA_MODULES_URL
+	mkdir -p $(dir $@)
+	curl -L "$(EXTRA_MODULES_URL)" > $@
+endif
+
 ifdef COMPILED_KERNEL_URL
 
 installer: minimal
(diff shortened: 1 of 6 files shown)


ibuildthecloud:
@tetatetit Sorry, but we just massively refactored the build.  We've been working on the v0.5 branch for awhile now but we just merged it to master.

bachp:
Is it now possible to auto format the state partition using `btrfs` and use the btrfs storage driver with it? I could not find anything in the documentation about it.

joshwget:
Sorry, didn't mean to close this!

@bachp Autoformat only supports ext4, so it's not currently possible.

PR #992, opened 2016-06-09
use latest arm64 kernel build

imikushin:
use latest arm64 kernel build

for v0.5 branch


ibuildthecloud:
@imikushin why did this build fail?

imikushin:
@ibuildthecloud no idea. just in case, I rebased it onto a later commit in v0.5 (that I believe to be building fine). let's see if it works.

PR #1011, opened 2016-06-14
Fix installer build

imikushin:
Fix installer build


joshwget:
The test here will verify that an installer image was successfully built, but it won't actually verify that it performs a successful installation. It'll just reboot into the ISO again, which doesn't actually check anything.

PR #1013, opened 2016-06-14
Additional mounts can be set in cloud config

joshwget:
Switched #923 to use the standard `mounts` key. Still no `swap` key support, but this can be added later.


deniseschannon:
https://github.com/rancher/os/issues/741

Issue #979, opened 2016-06-03
/etc/selinux/ros/seusers does not upgrade when we upgrade RancherOS


datawolf:
**RancherOS Version: (ros os version)**
Upgrade from 0.4.4 to 0.4.5

**Where are you running RancherOS? (docker-machine, AWS, GCE, baremetal, etc.)**
baremetal

on RancherOS v0.4.4, the content of the file `/etc/selinux/ros/seusers` is

```
[rancher@rancher-dev ~]$ cat /etc/selinux/ros/seusers 
__default__:system_u:s0-s0
```
with the commit https://github.com/rancher/os/commit/3fa11f963d3989c229f6c97924354ab8f6547407, the content of this file is updated.

but when I use `sudo ros os upgrade -i rancher/os:v0.4.5` to upgrade to RancherOS v0.4.5. and reboot the OS. I found that the content of the file `/etc/selinux/ros/seusers`  has not been update.

```
[rancher@rancher-dev ~]$ cat /etc/selinux/ros/seusers 
__default__:system_u:s0-s0
```

I think it should be 

```
[rancher@rancher-dev ~]$ cat /etc/selinux/ros/seusers 
__default__:unconfined_u:s0-s0:c0.c1023
```

the file `/etc/selinux/ros/contexts/failsafe_context` has the same problem.

joshwget:
Sorry, this was certainly an oversight with the v0.4.5 release. Unfortunately, we won't be able to provide a fix for this until v0.5 since the v0.4.5 installer has already been released. We'll be sure that all SELinux files are updated as part of the v0.5 installer.

joshwget:
Confirmed with rancher/os:v0.5.0-rc1

PR #1022, opened 2016-06-17
Update SELinux config files in installer

joshwget:
#979

### DIFF
--- a/scripts/installer/lay-down-os
+++ b/scripts/installer/lay-down-os
@@ -194,6 +194,15 @@ grub2_config "${KERNEL_ARGS}"
 pvgrub_config "${KERNEL_ARGS}"
 install_rancher
 
+seusers=${BASE_DIR}/etc/selinux/ros/seusers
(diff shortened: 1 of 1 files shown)


scripts/installer/lay-down-os
```diff
@@ -194,6 +194,9 @@ grub2_config "${KERNEL_ARGS}"
 pvgrub_config "${KERNEL_ARGS}"
 install_rancher
 
+echo "__default__:unconfined_u:s0-s0:c0.c1023" > ${BASE_DIR}/etc/selinux/ros/seusers
```
ibuildthecloud: Can you think of a situation in which `/etc/selinux/ros/` might not exists?  Maybe you should `mkdir -p /etc/selinux/ros/context` just in case.

joshwget: I added a check to see if the file exists. I think the problem was just that the file doesn't exist on first install.

PR #1026, opened 2016-06-17
Make rancher.debug control bootstrap Docker and turn on log

ibuildthecloud:


config/disk.go
```diff
@@ -99,6 +99,10 @@ func applyDebugFlags(rawCfg map[interface{}]interface{}) map[interface{}]interfa
 		if !util.Contains(cfg.Rancher.SystemDocker.Args, "-D") {
 			cfg.Rancher.SystemDocker.Args = append(cfg.Rancher.SystemDocker.Args, "-D")
 		}
+		if !util.Contains(cfg.Rancher.BootstrapDocker.Args, "-D") {
+			cfg.Rancher.BootstrapDocker.Args = append(cfg.Rancher.BootstrapDocker.Args, "-D")
+		}
+		cfg.Rancher.Log = true
```
joshwget: This won't actually change the value. You'll have to change `rawCfg` like the couple lines below this.

ibuildthecloud: Thanks, will change

Issue #670, opened 2015-11-30
Default password remains after ISO install


robbertkl:
I've recently installed RancherOS (v0.4.1) in 2 different ways:

1. After booting from the ISO
2. After booting with iPXE, passing `rancher.password=rancher` to the kernel

In both cases, I've used username *rancher* + password *rancher* to do the install: scp my cloud-config.yml containing my SSH pubkey, then run `sudo ros install -c cloud-config.yml -d /dev/sda`.

After the install, I was able to login as user *rancher* with my SSH key. However, after disabling my SSH key, I've discovered that in the 1st case (installed with ISO) I could still login with password *rancher*!

Checking `/etc/shadow`, the one installed from ISO indeed contains a password hash for user *rancher*, while the one installed from iPXE contains `!`, blocking password login. Both were installed with `ros install` and both had a user *rancher* with password *rancher* at the time of running the install (either baked in in the ISO, or set through a kernel parameter).

Seems like a serious security issue: a default password set on installed servers, while users unaware of this, since they're using SSH keys and expecting password authentication to be locked.

imikushin:
@robbertkl you must have rebooted from the ISO and it has set the password in the console container. 

If you're using an ephemeral (default) console, that should be fine: just remove the ISO and boot from your disk. With a persistent console (debian or ubuntu) the changes to /etc/shadow are saved and will persist across reboots.

robbertkl:
This was indeed the issue, the ISO was still attached. What threw me off here was that I was able to log in using my SSH key (given during installation), making me think it booted from disk. I guess this is because the ISO still detected my installed RancherOS because of the disk label `RANCHER_STATE`?

Removing the ISO does fix the issue (in case of the default console), but - like you explain above - the password does remain on a persistent console. Does this mean if you've ever booted from ISO once (with a persistent console), the password will remain there forever? (if not manually removed)

Is there some documentation on what exactly is persisted on a ephemeral vs. persistent console? For example:
* What is always persisted, even on a ephemeral console? (home dirs, `/mnt`, `/opt`?)
* What is only persisted on persistent consoles? (`/usr/local`, `/etc`?)
* What is never persisted, even on a persistent console? (anything at all?)

joshwget:
You're correct, the problem here is that the ISO will use the `RANCHER_STATE` disk by default. This can interfere with existing installations, so it's definitely not a very good default behavior. We're going to change this in v0.5.1.

joshwget:
[...]

Issue #662, opened 2015-11-25
It seems rancher os cannot be recovered from a rancher os stick


blaggacao:
After corrupting my cloud config yaml, I wanted to recover from the initial rancher os stick, with wich I infected the computer on the first hand. Aparently, the stick also reads the cloud_config file from disk, when in place and therefore fails.

Not 100% sure of this interpretation of what I see and not sure if it's intended.
https://github.com/rancher/rancher/issues/2803#issuecomment-159490706

ibuildthecloud:
@blaggacao You can boot rancheros with the kernel command line "rancher.state.dev=none". That will prevent it from using the state partition. You can then mount the state partition from the live system and recover.

blaggacao:
Whohoo! Can we put the label documentation on that? When someone's working with bare metal, that question might pop up every so often...

2devnull:
what credentials are used with this method? It is not authenticating with ssh key nor rancher/rancher.

ibuildthecloud:
The password of the rancher user is controlled with another cmdline param  `rancher.password=rancher`

ibuildthecloud:
We should really just make a "rescue" option in the bootable USB/ISO

blaggacao:
@ibuildthecloud That's would be convenient, I know you go for convenience! So yeay!

ibuildthecloud:
@blaggacao Contributions welcome :) Our syslinux.cfg is available at https://github.com/rancher/os/blob/master/scripts/isolinux.cfg.  It's barebones, we really need a more robust one that has a bit of a pause to allow you to type things on the command line.  Right now it just boots straight it, would be nice to display a message and some key like "F2 Rescue."  Like you see on other distros' ISOs.

2devnull:
I don't see that when I boot. I only see the GRUB2 selection. Perhaps it can be added there.

ibuildthecloud:
@2devnull That's a good point we should add this to boot the GRUB2 boot loader and the syslinux one we use from the ISO/USB.  The grub2 config is controlled from https://github.com/rancher/os-installer/blob/master/scripts/lay-down-os#L72

joshwget:
Confirmed with v0.6.0-rc3 that the ISO doesn't use the state partition by default, which prevents this issue.

PR #1034, opened 2016-06-19
Don't mount RANCHER_STATE by default when booting off ISO

joshwget:
#670 
#662


joshwget:
Closing in favor of #1128

PR #1042, opened 2016-06-22
Add swap support in cloud config mounts

joshwget:


scripts/run
```diff
@@ -138,6 +141,12 @@ if [ "$QEMU" == "1" ]; then
         qemu-img create -f qcow2 -o size=10G ${HD}
     fi
 
+    HD2=${BASE}/state/hd2.img
+    if [ "$SECOND_DRIVE" == "1" ]; then
+        qemu-img create -f qcow2 -o size=10G ${HD2}
```
ibuildthecloud: cool feature, but can we respect the `$FRESH` flag?  So if it exists we use it otherwise if `$FRESH == 1` we delete it.

joshwget: It definitely should, I'll update

PR #1047, opened 2016-06-27
Fix multi-arch support for releases

datawolf:


joshwget:
LGTM @ibuildthecloud @imikushin

Issue #1065, opened 2016-06-29
SSH keys in a GCE project are no longer automatically being passed into a GCE instance


deniseschannon:
**RancherOS Version: (ros os version)**
v0.5.0-rc1

**Where are you running RancherOS? (docker-machine, AWS, GCE, baremetal, etc.)**
GCE

Using a GCE image for v0.5.0-rc1, the ssh keys in my GCE project are not being automatically added to my OS, so I cannot log in unless I pass in a cloud-config with my SSH keys.

deniseschannon:
Verified with v0.5.0-rc3. 

Able to log into a GCE instance based on SSH keys inside the GCE project. No keys need to be passed in using cloud-config.

PR #1078, opened 2016-06-30
Do not lose ssh keys from metadata if there are no ssh keys in cloud-config

imikushin:
Do not lose ssh keys from metadata if there are no ssh keys in cloud-config


deniseschannon:
https://github.com/rancher/os/issues/1065

Issue #1118, opened 2016-07-18
cloud-config option to set NTP time servers


joshwget:

joshwget:
Confirmed with v0.6.0-rc3 that `write_files` and `restart_services` can be used to set NTP time servers in `/etc/ntp.conf`.

PR #1119, opened 2016-07-19
Add rsyslog configuration to system-volumes.

allingeek:
This change enables a user to inject rsyslog config via cloud-config. Added these two bind-mount volumes to make the ROS rsyslog config accessible to the syslog system service.

  - /etc/rsyslog.conf:/etc/rsyslog.conf
  - /etc/rsyslog.d:/etc/rsyslog.d

### DIFF
--- a/os-config.tpl.yml
+++ b/os-config.tpl.yml
@@ -285,6 +285,8 @@ rancher:
       - /etc/hosts:/etc/hosts
       - /etc/resolv.conf:/etc/resolv.conf
       - /etc/ssl/certs/ca-certificates.crt:/etc/ssl/certs/ca-certificates.crt.rancher
+      - /etc/rsyslog.conf:/etc/rsyslog.conf
+      - /etc/rsyslog.d:/etc/rsyslog.d
       - /etc/selinux:/etc/selinux
       - /lib/firmware:/lib/firmware
       - /lib/modules:/lib/modules


allingeek:
I've yet to confirm, but I believe that this will require that the standard config is placed in the base FS instead of the syslog image. I'm open to other solutions.

ibuildthecloud:
@allingeek We have a very similar problem with ntp also.  We were thinking of a generic solution in which we have a configuration syntax in cloud-config like:

```yaml
#cloud-config
rancher:
  files:
    ntp.conf: |
      contents ...
    rsyslog.conf: |
      contents ...
```

We would then change the ntp, rsyslog, or any other container to then look for these files in the cloud-config and use them.  We first started with the idea of using `write_files` in cloud config but we have many initialiization issues because write_files typically runs too late.  One down side of the approach that I just described is that it looks a lot like write_files, but is different in function (think pull not push).  So that is a bit confusing to a user the subtly.

@joshwget

allingeek:
I like the pull vs. push model. I'm not worried about special casing the default services because there really aren't that many of them. 

Another alternative would be to specify images to use for default system services with cloud-config. Something like:

    #cloud-config
    rancher:
      syslog-image: allingeek/syslog

Building a whole image just to inject config might feel like a burden at first, but at least it would give everyone a nice simple interface to agree on.

joshwget:
Using custom images will still be available even if we add an option to pull in configuration files from cloud-config. It's likely the former will always be needed for some use cases, but the latter has less overhead for the vast majority of cases.

allingeek:
Sounds great. Is there another issue tracking that improvement? I'd like to watch that and close this one.

joshwget:
It's sort of tracked in #1118, though there isn't an issue tracking a more general solution like what was proposed here.

ryansch:
Is there a way to accomplish this in the current 0.5.0 release?  I can't figure out how to replace the syslog container so I can push logs to papertrail.

allingeek:
@ryansch If you're talking about application logs, @cloudnautique put together a great [tutorial](http://rancher.com/aggregating-application-logs-from-docker-containers-with-loggly/) with syslog forwarding (to Loggly). It uses an "application" level rsyslog container and exposes the syslog socket via volumes. Applications can log directly to that socket if the syslog container is a volumes-from dependency.

ryansch:
Thanks for the reply @allingeek.  I'm using logspout in that use case.  Due to regulatory concerns I need to persist my ssh logs in a centralized location.  I figured out how to replace the normal syslog container with one that will forward to papertrail.

joshwget:
Thanks for the PR @allingeek. As discussed, we're going to go with the method in #1147.

Issue #850, opened 2016-04-01
[os-images] os-udev has a bug?


datawolf:
When install RancherOS into disk (eg /dev/sda) on x86 server(24 cpu cores). and reboot the RancherOS, it can not mount /dev/sda1.

I debug the os start and found that the `ResoloveDevice(LABEL=RANCHER_STATE)` failed. 
but when I set `maxcpus=1` in kernel command line, the `ResoloveDevice(LABEL=RANCHER_STATE)` can run successfully.

I think there may be some bug in container `udev-bootstrap(image: os-udev)` on multiple processor server.

**The issue is very serious, because we can not install RancherOS into the multiple processor server besides setting the `maxcpus=1` in kernel command line.**

sheng-liang:
Can you include the output of udev.  do `sudo system-docker logs udev` also the content of `/var/log/messages

ibuildthecloud:
@datawolf I don't think there is any fundamental issue w/ multi-core as we always run RancherOS on multicore systems, but there could be something with your particular setup that is causing issues.  There could possible be an issue in the udev doesn't settle fast enough.  I'm pretty sure we wait, but maybe there is a bug in which the driver has yet to load before we check for the device.  After you boot with all cores enabled does `/dev/sda1` exist?

datawolf:
@sheng-liang The output of udev is empty.

datawolf:
@ibuildthecloud You are right, the driver has yet to load before RancherOS check for the device.

ibuildthecloud:
If you know the module that must be loaded you can try adding  to the kernel command line `rancher.modules=[ata_piix]`, that example will load ata_piix module.

ibuildthecloud:
We will investigate the udev settle issue.

datawolf:
@ibuildthecloud I add `rancher.modules=[mpt3sas0]` to the kernel command line, the problem is still exist. Although the driver is loading before the check, it need about 20s to load OK. so RancherOS check for the device will failed.

```
$ dmesg |grep mpt3sas0 
[ 10.273367] mpt3sas0: 64 BIT PCI BUS DMA ADDRESSING SUPPORTED, total mem (65647196 kB)
[ 10.328432] mpt3sas0: MSI-X vectors supported: 8, no of cores: 12, max_msix_vectors: 8
[ 10.329004] mpt3sas0-msix0: PCI-MSI-X enabled: IRQ 37
[...]

PR #1128, opened 2016-07-25
Refactor state

joshwget:
Sets the ISO to no longer use the `RANCHER_STATE` drive by default. Adds a `rancher.root` cloud-config key to configure the root drive. Whenever this key is specified, wait for the root drive to be initialized.

#670 
#662 
#850

### DIFF
--- a/images/02-autoformat/auto-format.sh
+++ b/images/02-autoformat/auto-format.sh
@@ -12,7 +12,7 @@ for dev in ${DEVS[@]}; do
 
         # Test for our magic string (it means that the disk was made by ./boot2docker init)
         HEADER=`dd if=${dev} bs=1 count=${#MAGIC} 2>/dev/null`
-    
+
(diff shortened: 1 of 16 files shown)


images/02-autoformat/auto-format.sh
```diff
@@ -23,10 +23,10 @@ for dev in ${DEVS[@]}; do
             # do not auto-format if the disk does not begin with 1MB filled with 00
             continue
         fi
-    
-        mkfs.ext4 -L RANCHER_STATE ${dev}
-    
+
+
         if [ -e "/userdata.tar" ]; then
+            mkfs.ext4 -L B2D_STATE ${dev}
```
PaulCharlton: @joshwget why did you choose a different label (B2D_STATE) when preserving the contents of the volume? I am missing some context and did not see it referenced in the issues.

PR #1136, opened 2016-07-29
WIP: Make it easier to switch between Docker engines

joshwget:


images/02-console/docker-init
```diff
@@ -11,10 +11,15 @@ done
 
 DOCKER_BIN=$(which docker) || DOCKER_BIN=/usr/bin/docker
 
-for i in /opt/bin /usr/local/bin; do
+for i in /opt/bin /usr/local/bin /var/lib/rancher/docker; do
+    if [ -x ${i}/dockerd ]; then
```
ibuildthecloud: Why do we need to copy these?  Can we not just add then to the PATH in the front?

ibuildthecloud: I'm concerned when we run in memory that each `cp` will take more and more memory.

os-config.tpl.yml
```diff
@@ -292,6 +288,7 @@ rancher:
       - /usr/share/ros:/usr/share/ros
       - /var/lib/rancher/cache:/var/lib/rancher/cache
       - /var/lib/rancher/conf:/var/lib/rancher/conf
+      - /var/lib/rancher/docker:/var/lib/rancher/docker
```
ibuildthecloud: I think these `/var/lib/rancher/*` are not needed because we bind mount `/var/lib/rancher`.  I think the only reason why we kept these was because some code expected these folders to exist and the bind mount does an implicit `mkdir`.  We should just fix that issue instead.

images/02-console/docker-init
joshwget: This is the biggest part that I was concerned about. Since the binaries are now supplied via the Docker container, there aren't any binaries in the console container until they're copied. I think most users expect them to be in `/usr/bin` even if they aren't directly launching the Docker daemon.

Another important piece of this is the client. We'd have to update the PATH in the console itself for the client to be located in `/var/lib/rancher/docker`.

os-config.tpl.yml
joshwget: I agree, these aren't really necessary.

Issue #1086, opened 2016-07-02
won't obey udev rule?


rafaelsoaresbr:
**RancherOS Version: v5.0**

**Where are you running RancherOS? baremetal**

I've created a file "write_files.yml":

```
write_files:
  - path: /etc/udev/rules.d/99-lowbat.rules 
    permissions: "0755"
    owner: root
    content: |
      SUBSYSTEM=="power_supply", ATTR{status}=="Discharging", RUN+="/usr/sbin/poweroff"
```

and meged it with:

`sudo ros config merge -i write_files.yml`

Then I've unplugged the power cable and nothing happened.

joshwget:
#1147 should allow you to write to this file in the udev container.

PR #1147, opened 2016-08-10
Extend write_files to run in all system services

joshwget:
Depends on #1145 and rancher/os-images#103

Adds a `container` key to `write_files` that allows writing to files in services other than `console`. This is done via a custom entrypoint. The existing entrypoint was rewritten in Go for this PR.

Adds a `rancher.restart_services` key that allows restarting a system service later in the boot process. This is especially useful for services that need to reapply `write_files`.

Supersedes #1119 
#1118


rafaelsoaresbr:
Can this solve this issue #1086 ?

joshwget:
@rafaelsoaresbr I'd have to try it out to be sure, but I think it should.

PR #1152, opened 2016-08-16
Strip more Docker binaries from DFS image

joshwget:
Remove some unnecessary Docker binaries that were included in the last RC. The final ISO size was bigger than it should have been as a result of this.


ibuildthecloud:
@joshwget Will `exclude=usr/bin/docker*` work?  You don't have `dockerd` in the list.

joshwget:
You're right, that's a better approach.

PR #1155, opened 2016-08-16
Correct permissions on home directory

joshwget:


cmd/console/console.go
```diff
@@ -29,18 +29,22 @@ type symlink struct {
...
-	if err := os.MkdirAll(dockerHome, 2755); err != nil {
-		log.Error(err)
-	}
-	if err := os.Chown(dockerHome, 1101, 1101); err != nil {
-		log.Error(err)
+	if _, err := os.Stat(dockerHome); os.IsNotExist(err) {
+		if err := os.MkdirAll(dockerHome, 0755); err != nil {
+			log.Error(err)
+		}
+		if err := os.Chown(dockerHome, 1100, 1100); err != nil {
```
ibuildthecloud: Is 1100 the right number here?  Before it was 1101

joshwget: No, these should be 1101. Updated.

Issue #787, opened 2016-02-27
ISO is (potentially) dangerous on hosts in the DMZ (edge servers)


adrianguenter:
To give you an idea of my setup, I'm running ROS on an "unmanaged" KVM (Qemu) VPS, which provides SolusVM for configuring/managing the instance. The "eth0" interface is bound to an IPv4 and IPv6/64, both in the DMZ. I have no access control over these networks – I can essentially only set reverse DNS for the IPv4.

The following combine to create a scenario where anyone can easily root a publicly accessible RancherOS host after it boots the ISO, just by knowing its hostname/IP and the default "rancher/rancher" user/pass combo, and then winning the race to establishing a shell via SSH:

- SSH starts automatically…
  - On port 22
  - Bound to all interfaces
  - With password authentication enabled
  - With `AllowGroups docker` (of which the "rancher" user is a member)
- User "rancher" is equivalent to root
  - /etc/sudoers contains `rancher ALL=(ALL) NOPASSWD: ALL`
  - `sudo whoami` → `root`
- User "rancher" has an insecure password which is…
  - Documented
  - Easily guessed, regardless

This compromises all data on any physically attached storage (e.g. the block storage attached to a VPS instance in my case), and can also lead to exploits baked into the new init image itself during the `ros install` process. The only way to safely (re)install ROS to disk in this environment is to win the race to `ssh rancher@host -t passwd` after sshd starts accepting connections. I personally wouldn't like my odds against a skilled adversary or their bot(s) under these conditions.

(i)PXE adds a layer of complexity to my setup that I would very much like to avoid. Moreover, the official iPXE ISO – the only one available on my VPS host – is intended for "experimental use" and built without HTTPS, as the iPXE team intends for you to bake your own image containing the specific CA cert(s) needed. As far as I can tell an official build of iPXE with HTTPS (containing a sane root certificate bundle) doesn't exist, and pulling ROS initrd over HTTP is obviously a step in the wrong direction.

Is there any reason why the "rancher" user's password can't be generated from the "urandom" device file (using something like [this accepted answer on ServerFault](http://serverfault.com/a/261417)) on boot, and then printed to the console right before the login prompt is issued? I have VNC access to my instance, so this would be sufficient for me personally.

There are probably use cases I'm overlooking (headless hardware without VNC or other access to the main virtual console/terminal?), but as it stands I'm nervous every time I boot the ISO. I'm hoping that a low-friction solution exists, and would be happy to contribute to its development, if possible.

Thanks!

[...]

PR #1157, opened 2016-08-16
Add rancher.autologin key

joshwget:
Depends on rancher/os-base#35

#787


ibuildthecloud:
os-base PR https://github.com/rancher/os/pull/1161

PR #1161, opened 2016-08-17
os-base v2016.05-5

ibuildthecloud:


ibuildthecloud:
Oh shoot just saw: https://github.com/rancher/os-base/pull/36  I'll update to v2016.05-5 after the build is done.

PR #1164, opened 2016-08-18
Remove -s overlay from rancher.docker.args as overlay is used by default

imikushin:
Remove -s overlay from rancher.docker.args as overlay is used by default

This change will make it easier to customize docker storage by adding arguments to rancher.docker.extra_args and not having to overwrite rancher.docker.args list.

### DIFF
--- a/os-config.tpl.yml
+++ b/os-config.tpl.yml
@@ -387,4 +387,4 @@ rancher:
     engine: docker-1.11.2
     tls_args: [--tlsverify, --tlscacert=/etc/docker/tls/ca.pem, --tlscert=/etc/docker/tls/server-cert.pem, --tlskey=/etc/docker/tls/server-key.pem,
       '-H=0.0.0.0:2376']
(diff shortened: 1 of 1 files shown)


ibuildthecloud:
@imikushin Do you know what version of Docker this change was made?  We are currently supporting 1.10.3, 1.11.2, 1.12.0.

imikushin:
I'm sure about docker-1.11, but not exactly sure about docker-1.10.

In ROS-v0.4.5 booted from ISO, docker doesn't start if `-s overlay` isn't in docker args. 

Perhaps, instead of removing it altogether, a better idea would be to move `-s overlay` into rancher.docker.extra_args. What do you think?

joshwget:
If we move it to `extra_args` it might get replaced accidentally. For example, a user runs `ros config set rancher.docker.extra_args [-g,/mnt/docker]` and without knowing it they changed the storage driver. `extra_args` probably serves its purpose best if it's empty by default.

If we can wait until the next release, #1166 is probably a better long-term solution.

joshwget:
@imikushin Are you okay with closing this? I think it's superseded by #1239.

ibuildthecloud:
Closing due to https://github.com/rancher/os/pull/1239

Issue #1102, opened 2016-07-07
Implement 'runcmd' cloud-config keys


joshwget:
`runcmd` should probably just be a simpler alternative to writing to `/opt/rancher/bin/start.sh`. This means that the commands will:

1. Run at the start of the console container
2. Run before User Docker begins
3. Be run on each boot (or each time the console container is restarted, more accurately)

I'm still unsure when `bootcmd` should be applied to maximize its usefulness.

drpebcak:
👍

deniseschannon:
Verified with v0.6.0-rc4

PR #1170, opened 2016-08-22
Implement runcmd

joshwget:
Provides an implementation of the cloud-init `runcmd` key. Currently only supports the list syntax, whereas cloud-init also provides the option of specifying commands as strings.

By design, RancherOS will run all of the commands on each boot. This differs from cloud-init which runs all of the commands only once.

#1102

### DIFF
--- a/cmd/cloudinitexecute/cloudinitexecute.go
+++ b/cmd/cloudinitexecute/cloudinitexecute.go
@@ -89,6 +89,20 @@ func ApplyConsole(cfg *rancherConfig.CloudConfig) {
 			log.Errorf("Failed to mount %s: %v", mount[1], err)
 		}
 	}
+
+	for _, runcmd := range cfg.Runcmd {
+		if len(runcmd) == 0 {
+			continue
+		}
+
+		cmd := exec.Command(runcmd[0], runcmd[1:]...)
+		cmd.Stdin = os.Stdin
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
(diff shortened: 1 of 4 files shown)


cmd/cloudinitexecute/cloudinitexecute.go
```diff
@@ -89,6 +89,22 @@ func ApplyConsole(cfg *rancherConfig.CloudConfig) {
 			log.Errorf("Failed to mount %s: %v", mount[1], err)
 		}
 	}
+
+	for _, runcmd := range cfg.Runcmd {
+		var cmd *exec.Cmd
+
+		if len(runcmd) == 0 {
+			continue
+		} else if len(runcmd) == 1 {
```
ibuildthecloud: FYI, this is redundant, if the length of a slice is X, then the slice `[X:]` is valid.  Only if you do `X+1` will you get a panic

cmd/cloudinitexecute/cloudinitexecute.go
```diff
@@ -89,6 +89,22 @@ func ApplyConsole(cfg *rancherConfig.CloudConfig) {
...
+			cmd = exec.Command(runcmd[0])
+		} else {
+			cmd = exec.Command(runcmd[0], runcmd[1:]...)
+		}
+
+		if err := cmd.Run(); err != nil {
```
ibuildthecloud: can you set `cmd.Stdin` and `cmd.Stdout` to `os.Stdin` and `os.Stdout`

PR #1172, opened 2016-08-22
Make respawn slightly extensible

ibuildthecloud:


cmd/console/console.go
```diff
@@ -172,7 +165,32 @@ func generateRespawnConf(cmdline string) []byte {
...
+	if err == nil {
+		for _, f := range files {
+			content, err := ioutil.ReadFile(f.Name())
+			if err != nil {
+				return err
+			}
+			respawn += fmt.Sprintf("\n%s", string(content))
+		}
+	} else if !os.IsNotExist(err) {
+		return err
```
joshwget: We probably don't want to return here. Still need to write `/etc/respawn.conf` even if there's an error reading `/etc/respawn.conf.d`.

Issue #1193, opened 2016-08-27
Can not  restart user-docker at second time?


datawolf:
**RancherOS Version: (ros os version)**

v0.5.0

**Where are you running RancherOS? (docker-machine, AWS, GCE, baremetal, etc.)**

qemu and packet.net

when I use command `system-docker restart docker`, the first time it can run success, but the second time , it runs failed.

it can be reproduced by the following steps:

```
[root@rancher rancher]# docker run -d ubuntu:ping bash -c "ping www.google.com"
498b79d4b3262143fb4878b0feecd90b1b3cdb61c8c9d01601c5729726c2b002
[root@rancher rancher]# docker ps
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS               NAMES
498b79d4b326        ubuntu:ping         "bash -c 'ping www.go"   3 seconds ago       Up 3 seconds                            silly_perlman
[root@rancher rancher]# system-docker restart docker
docker
[root@rancher rancher]# docker ps
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
[root@rancher rancher]# system-docker restart docker
docker
[root@rancher rancher]# docker ps
An error occurred trying to connect: Get http://%2Fvar%2Frun%2Fdocker.sock/v1.23/containers/json: read unix @->/var/run/docker.sock: read: connection reset by peer
[root@rancher rancher]# tail -f /var/log/docker.log
time="2016-08-27T01:49:43.751578855Z" level=info msg="previous instance of containerd still alive (1889)"
time="2016-08-27T01:49:44.778426808Z" level=fatal msg="Error starting daemon: timeout"
time="2016-08-27T01:49:46.210952543Z" level=info msg="previous instance of containerd still alive (1889)"
time="2016-08-27T01:49:47.233005236Z" level=fatal msg="Error starting daemon: timeout"
time="2016-08-27T01:49:49.068177896Z" level=info msg="previous instance of containerd still alive (1889)"
time="2016-08-27T01:49:50.090340478Z" level=fatal msg="Error starting daemon: timeout"
time="2016-08-27T01:49:52.710271848Z" level=info msg="previous instance of containerd still alive (1889)"
time="2016-08-27T01:49:53.731917726Z" level=fatal msg="Error starting daemon: timeout"
time="2016-08-27T01:49:57.970449681Z" level=info msg="previous instance of containerd still alive (1889)"
[...]

PR #1220, opened 2016-09-07
Use runc exec to start User Docker

joshwget:
Depends on rancher/os-services#49

#1193

### DIFF
--- a/cmd/userdocker/main.go
+++ b/cmd/userdocker/main.go
@@ -4,9 +4,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
-	"os/signal"
 	"path"
-	"strconv"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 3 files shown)


cmd/userdocker/main.go
```diff
@@ -191,73 +167,23 @@ func startDocker(cfg *config.CloudConfig) (string, types.HijackedResponse, error
...
-		if os.IsNotExist(err) {
-			log.Infof("Waiting for %s", file)
-			time.Sleep(1 * time.Second)
-		} else if err != nil {
-			return nil, err
-		} else {
-			return contents, nil
-		}
-	}
+	return syscall.Exec("/usr/bin/docker-runc", cmd, os.Environ())
```
ibuildthecloud: If you are doing `syscall.Exec` then you don't need to symlink.  Just change to `syscall.Exec("/usr/bin/ros", cmd, os.Environ())`

PR #1236, opened 2016-09-14
Refactor how consoles and engines are loaded

joshwget:


cmd/switchconsole/switch_console.go
```diff
@@ -25,9 +24,7 @@ func Main() {
 	}
 
 	if newConsole != "default" {
-                project.ServiceConfigs.Add("console", &composeConfig.ServiceConfig{})
-
-		if err = compose.LoadService(project, cfg, true, newConsole); err != nil {
+		if err = compose.LoadSpecialService(project, cfg, "console", newConsole); err != nil {
```
ibuildthecloud: Now the other service don't think they're special.

joshwget: They had to find out sometime :cry:

PR #1167, opened 2016-08-19
Bump Go to 1.7.1

joshwget:

PR #1240, opened 2016-09-15
Enable scripts/validate without golint

joshwget:
Depends on #1167


ibuildthecloud:
What's wrong with golint?

joshwget:
Nothing, just haven't taken the time to go through and fix all of the golint errors. All three are currently disabled, so I'd rather have gofmt and vet than nothing.

PR #1260, opened 2016-09-30
Add noformat and RAID installer types

joshwget:


scripts/installer/lay-down-os
```diff
@@ -69,15 +80,29 @@ mount_device()
...
+}
+
+install_syslinux() {
+    dd bs=440 count=1 if=/usr/lib/syslinux/mbr/mbr.bin of=${DEVICE}
+    cp /usr/lib/syslinux/modules/bios/* ${BASE_DIR}/${BOOT}syslinux
+    extlinux --install ${BASE_DIR}/${BOOT}syslinux
+}
+
+install_syslinux_raid() {
+    dd bs=440 count=1 if=/usr/lib/syslinux/mbr/mbr.bin of=/dev/sda
```
ibuildthecloud: This can be done in a future PR but we really should determine the devices that are backing ${DEVICE} and not hardcode `/dev/sd{a,b}`

Issue #1237, opened 2016-09-15
FS Resize Unreliable


ryansch:
**RancherOS Version: (ros os version)**
0.6.1
**Where are you running RancherOS? (docker-machine, AWS, GCE, baremetal, etc.)**
AWS

I started 9 instances and only 5 of them did the resize.  I manually ran the 0.4.5 resize container on the other 4 and it did the right thing.

These instances are started through two different ASGs so they should be identical within those two groups.  Both groups were affected.

joshwget:
Can you post the output of `system-docker logs cloud-init-execute` for one of the instances where resize failed?

ryansch:
```
time="2016-09-14T17:45:29Z" level=info msg="Running cloud-init-execute: pre-console=true, console=false"
time="2016-09-14T17:46:04Z" level=error msg="Failed to resize /dev/xvda: exit status 1"
```

ryansch:
I have `RESIZE_DEV: /dev/xvda` set in the cloud config environment.

joshwget:
Looks like we don't capture the output from the resize commands, so the logs don't reveal too much. I'll spin up some instances until I can reproduce this and then debug from there. Thanks @ryansch!

ryansch:
Thanks!

schuylr:
Seeing this issue with v0.6.0 as well.

fifthecho:
I'm having this issue with ami-a8d2a4bf which is also preventing any further services or cloud-init data from executing.

joshwget:
The following can be used as a workaround.

```
rancher:
  environment:
    RESIZE_DEV: /dev/xvda
  services_include:
    https://raw.githubusercontent.com/rancher/os-services/v0.4.5/r/resize-fs.yml: true
```

schuylr:
Yep that hack works for me, I can also confirm that `docker run -it --rm --privileged slamper/disk_resizer xvda` (stolen from @CumpsD in #1099) will resize already-initialized instances.

deniseschannon:
Verified with v0.7.0-rc3. 

Ran 30 instances in AWS and all re-sized as expected.

PR #1299, opened 2016-10-08
Remove read-only mounts from cloud-init-execute

joshwget:
`resize2fs` looks at mount points for the device it's trying to resize. If the first one it finds is a read-only mount then it will fail with `resize2fs: Read-only file system While checking for on-line resizing support`.

#1237


ibuildthecloud:
@joshwget You have ninja like skills.

PR #1306, opened 2016-10-10
Revert "Add EFI live boot support"

ibuildthecloud:
Reverts rancher/os#1263


ibuildthecloud:
@deitch I'm sorry, this just seems plain mean, but this is blocking the 0.7.0 release.  I'm queuing this PR up just in case we need to release and we haven't gotten EFI for ARM fixed.  More detail here https://github.com/rancher/os/pull/1263#issuecomment-252753554

deitch:
> this just seems plain mean

Nah, no worries. Let's clean it up therein #1263

Issue #1287, opened 2016-10-05
Service commands shouldn't be direct copies from libcompose


joshwget:
Other than `enable`, `disable`, `list`, and `delete`, all `ros service` commands are directly imported from the libcompose `cli` package. It would be better to have our own versions of these commands so we can tailor them to RancherOS use cases.

There are likely others, but these changes would be a good start.
- `-d` should be the default
- `-f` and `-p` should be disabled
- If no service is specified for an operation, we should throw an error rather than applying that operation to all services

deniseschannon:
Verified that the help looks fixed so that irrelevant commands are hidden in v0.8.0-rc4

PR #1343, opened 2016-10-22
Switch from imported libcompose service commands to customized ones

joshwget:
There are still some improvements that can be made, but this at least lays down a foundation and takes care of the following.
- `-d` is now the default and `--foreground` must be used to log. `--log` might actually be better for this.
- `-f` and `-p` are disabled.
- At least one service must be specified for all service subcommands other than `ps`.


ibuildthecloud:
I think we still need further discussion on this.  Like `build`?

joshwget:
We do, but I figured we could start here at least. I don't think there's really any story for `build` yet, so it should probably just be removed for now.

ibuildthecloud:
Definitely a good start.  Would like to say LGTM but drone is having problems.

ibuildthecloud:
Merging with the caveat the more work needs to be done to cleanup `ros service`

joshwget:
#1287

PR #1370, opened 2016-11-04
Add initial DigitalOcean Debian->RancherOS cloud-config.yml

klausenbusk:
See: https://github.com/rancher/rancher.github.io/pull/532

I have removed the unneeded echo line, but everything else it the same..


joshwget:
[LGTM!](https://github.com/rancher/rancher.github.io/pull/532#issuecomment-258483379)

PR #1439, opened 2016-12-01
Kernel args shellwords

tetatetit:
To allow whitespaces in rancher/cloud-config values.. Quotes double and single handled properly like in shell. Very important for network or local diskless/stateless booting. Tested in production as well as 'test' and 'integration-test' passed.


config/data_funcs.go
```diff
@@ -156,13 +157,17 @@ func unmarshalOrReturnString(value string) (result interface{}) {
 
 func parseCmdline(cmdLine string) map[interface{}]interface{} {
 	result := make(map[interface{}]interface{})
+	params, err := shellwords.Parse(cmdLine)
```
joshwget: Can you switch this to not use the global parser and explicitly set `ParseEnv` and `ParseBackticks` to false? If the global defaults were ever changed in go-shellwords it might cause some bad results.

```
p := shellwords.NewParser()
p.ParseEnv = false
p.ParseEnv = false
params, err := p.Parse(cmdLine)
```

vendor/github.com/mattn/go-shellwords/.travis.yml
```diff
@@ -2,7 +2,6 @@ language: go
 go:
   - tip
 before_install:
-  - go get github.com/axw/gocov/gocov
```
joshwget: Do you know why this change was included? When I run trash this line gets added back. It's not a big deal either way since it'll just get changed back whenever the vendor folder is updated again.

tetatetit: Because  https://github.com/rancher/os/blob/master/trash.conf contains `github.com/mattn/go-shellwords v1.0.0`  line 32 which updates https://github.com/tetatetit/os/tree/master/vendor/github.com/mattn/go-shellwords on each `trash` run to the latest https://github.com/mattn/go-shellwords where https://github.com/mattn/go-shellwords/blob/master/.travis.yml does not contain `- go get github.com/axw/gocov/gocov` appropriately. So I not sure why to keep this line in our repo while it is not in original repo from where trash updates it.

joshwget: trash will only updates based on the version in `trash.conf`. It's v1.0.0 for go-shellwords, so this line should remain (https://github.com/mattn/go-shellwords/blob/v1.0.0/.travis.yml#L5).

config/data_funcs.go
tetatetit: Did you mean?

```
p := shellwords.NewParser()
p.ParseEnv = false
p.ParseBacktick = false
params, err := p.Parse(cmdLine)
```
Since you duplicated same line twice
```
p.ParseEnv = false
p.ParseEnv = false
```

Yes, it makes sence, corrected

vendor/github.com/mattn/go-shellwords/.travis.yml
tetatetit: Ah, I did not notice that v1.0.0 differs by this line, sorry for my inattention. But in this case there is some bug in trash IMO since it takes master go-shellwords while should v1.0.0 accordingly to `trash.conf`

tetatetit: But do we really need v1.0.0 , not master. Otherwise we can just comment https://github.com/rancher/os/blob/master/trash.conf#L32 if trash supports commenting, otherwise to just remove. Since I don't think v1.0.0 tag of go-shellwords repo will be ever updated IMO

joshwget: It does seem like there is some bug in trash. Perhaps you're using an older version.

Using a tag that doesn't change is actually intended. The results of running trash should be deterministic and not depend on the master branch of a repo. To update go-shellwords in the future we'd have to update `trash.conf` to point to a new tag.

tetatetit: > Perhaps you're using an older version

Old version of trash? Just removed and updated via go get. And don't you mention:

> When I run trash this line gets added back

i.e. don't you mean that you have same situation locally that it takes master instead v1.0.0?

Also I did neither touch trash.conf nor run `trash` at all

joshwget: Do you know how this file changed if you never actually ran `trash`? In any case, I don't think it's a big deal if we leave this change in the PR (we can always change it back in follow up PR). I'm mainly curious how this change ended up being included.

tetatetit: I finally figured out what was the quirk was. I just harried up in that time to add this feature to at least to repo and build used in my projects so I just known nothing about trash in that time, was  enough inattentive so did not realize that shellwords-go is already in upstream repo so just downloaded and put it in my repo somehow did not realizing it is already there. So I put it from master, not from v1.0.0. I am sorry again for so long consuion and taking your time. Now I removed this change in this PR

tetatetit:
Our past conversation has dissapeared I see right aftar I  I updated this PR to  remove that weird difference of .travis.yml. Did you receive my last explanation why this difference happened I figured out? And is any more action is expected from side to get my PR merged?

joshwget:
Yeah, the history is still viewable on Github. Thank for for resolving that!

I've only done some basic testing so far, but I've found some issues with using go-shellwords.

There are some weird results due to the fact that we use a YAML parser to parse each parameter. As an example, consider `rancher.disable=[""]` as a kernel parameter. This will currently result in a list containing a single empty string. After this PR the quotes will be stripped away and it will result in an empty list.

The bigger problem is with boot loaders such as GRUB that preprocess kernel parameters. GRUB will already process quotes in pretty much the same way as go-shellwords. This means quotes will probably be stripped off twice, though I'm still testing this part.

tetatetit:
Hah, but e.g. `ros config set bootcmd [""] set bootcmd [""]` results exactly same empty list, i.e. 'ros config get bootcmd' results `[]` after it.  So with shellwords isn't cc/rancher kernel args are consistent with `ros config ` while without shellwords it is not?

joshwget:
@tetatetit Let's evaluate this again once YAML is out of the picture for kernel parameters (#1494). Right now there are too many components where quotes have special meaning and it's making it difficult to understand and test the possible scenarios.

PR #1465, opened 2016-12-14
LUKS/cryptsetup with SSH unlock, LVM, NBD, BTRFS, XFS

tetatetit:
all for state partiton fully supported. Everything working perfectly in production in my projects. Just did not test NBD yet although going to soon.

### DIFF
--- a/cmd/control/bootstrap.go
+++ b/cmd/control/bootstrap.go
@@ -22,9 +22,28 @@ func bootstrapAction(c *cli.Context) error {
 	cfg := config.LoadConfig()
 
 	if cfg.Rancher.State.MdadmScan {
-		if err := mdadmScan(); err != nil {
-			log.Errorf("Failed to run mdadm scan: %v", err)
+		cmdRun("mdadm", "--assemble", "--scan")
+	}
+
+	if cfg.Rancher.State.NetConf || cfg.Rancher.State.Nbd.Host != "" || cfg.Rancher.State.Decrypt {
+                if err := cmdStart("netconf"); err != nil {
+                        log.Errorf("Failed to start netconf: %v", err)
+                }
+	}
+
+	if cfg.Rancher.State.Nbd.Host != "" {
+		nbd := cfg.Rancher.State.Nbd
+		args := []string { nbd.Host }
+		if nbd.Port != 0 {
+			args = append(args, string(nbd.Port))
 		}
+		if nbd.Name != "" {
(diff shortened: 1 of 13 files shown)


joshwget:
@tetatetit Would you mind splitting this into two PRs, one for NBD/encryption and one for the rest? It looks like the packages necessary for NBD and encryption don't add much to the base image. The code for these parts look pretty good to me as well.

We can evaluate support for other filesystems seperately, especially since some of the code overlaps with what @SvenDowideit is working on for his installer refactoring.

tetatetit:
Actually NBD/encryption logic is coupled with LVM scan/initialization logic as each of them is related to storage volumes (block devices) initialization while BTRFS/XFS code is related to extra fs support (including installer touch). So I would rather split NBD/LVM/encrption and BTRFS/XFS to different PRs so it should not overlap with @SvenDowideit work on installer. Although ideally I would suggest just to split only changes to installer (related to BTRFS/XFS only) to different PR, while everything rest to leave in this PR

tetatetit:
Regarding splitting only installer - let me make note that Rancher OS is actually self installable without installer but just by specifying state device in cc.rancher.state.dev in kernel args or in oem-config.yml in oem labeled partition or mouned iso. So I would not mix installer with such kind of deployment which I successfuly used even on running Windows server with neither pyhsical access to it nor even serial console/kvm and which can be even automated scriptable (I have an idea of project of automated remote deployment of Rancher OS in such way to running Linux and Windows machines regardless physical or virtuall regarless of provider and even regardless does it have serial console/kvm or not, but all cames down to just providing key or credentials for remote access to machine and everything else is done automatically by scripts - i.e. installing isolinux, adding boot entry, rebooting already into Rancher OS, then paritigoining accordingly to script or configuration file initially provided and rebooting Rancher OS finally to use state partitiong already)

joshwget:
@tetatetit Separating NBD/LVM/encryption out from the rest is pretty much what I was thinking. It seems like a pretty good separation point between this and adding support for other filesystems.

We intentionally keep the installation/upgrade process pretty simple behind the scenes for scenarios like yours. Installation and upgrade are essentially just laying down a kernel and initrd. `ros install` and `ros os upgrade` are just one implementation of this process.

denji:
@joshwget This pull request will be prepared for the next pre-release/release version?

tetatetit:
I'll try, although lack a time a bit for now as contributor

PR #1497, opened 2017-01-05
Add docs for using custom Docker engines

joshwget:


deniseschannon:
LGTM, if this works with v0.7.1, can we add the docs to the main repo? @joshwget

joshwget:
Yeah, this would actually be better off there.

joshwget:
rancher/rancher.github.io#608

PR #2114, opened 2017-09-19
Replace Sirupsen/logrus package with rancher/os/log

juliengk:
Replace Sirupsen/logrus package with rancher/os/log


niusmallnan:
LGTM

Issue #2141, opened 2017-10-26
Some configurations in cloud-init are skipped if network is static configured 


liyimeng:
**RancherOS Version: (ros os version)**
v1.1.0
**Where are you running RancherOS? (docker-machine, AWS, GCE, baremetal, etc.)**
KVM 

I iPXE the node with the config as below. 
<pre>
EXTRA_CMDLINE: /init
hostname: 10-168-122-198
rancher:
  autologin: tty1
  cloud_init:
    datasources:
    - url:http://10.168.122.36:8888/cloudconfig?arch=x86_64&dc=dc1&dns=10.168.122.1&gw=10.168.122.1&ip=10.168.122.198&mac=52%3A54%3A00%3A71%3A99%3Aef&netmask=255.255.255.0&static=true&uuid=bb42b403-b435-4081-aab8-87b59b8b70a2
  console: ubuntu
  docker:
    engine: docker-1.12.6
  environment:
    EXTRA_CMDLINE: /init
  network:
    dns:
      nameservers:
      - 10.168.122.1
    interfaces:
      mac=52:54:00:71:99:ef:
        address: 10.168.122.198/24
        dhcp: false
        gateway: 10.168.122.1
  services_include:
    kvm-vm-tools: true
  state:
    autoformat:
    - /dev/vda
    dev: LABEL=RANCHER_STATE
ssh_authorized_keys:
- ssh-rsa "Removed private info"
</pre>

For some reason, my dns config can not be used during system start, and result other configuration ignored, e.g. docker version and console is not set as configured. Below is a screen shot which I suspect might result in the errors. 

![screenshot_ros-3_2017-10-26_15 34 43](https://user-images.githubusercontent.com/3155688/32055893-6246d480-ba63-11e7-9db8-1f17139d10f5.png)

liyimeng:
After a some troubleshotings, I can narrow down the issue now. If I try to set the dns, I have the problem, even I set it to 8.8.8.8. 
If I skip dns settings and lets rancheros pickup it is own, it works. 

Please note that, when I set dns, everything actually working OK after system start. I can use the dns server in the console. It just be ignored during system startup.

liyimeng:
This is likely a timing issue, more complicated cloud config file, more easy to get hit with the problem. 
For some reason, if the network is static configured, the interface is not get up and ready for system boot up.

liyimeng:
I close this one since it should not work at all with my setup. I properly have miss configuration somewhere on the LAN.

PR #2145, opened 2017-10-30
Ensure there is at least one link up before moving on

liyimeng:
Try to fix issue https://github.com/rancher/os/issues/2141

WIP: Code not tested because I have not proper setup at the moment.


liyimeng:
this change is unlikely necessary, I close it for now.

PR #2189, opened 2017-12-23
Fix user-data for ec2 metadata source

r0b2g1t:
The commit change the endpoint url of the ec2 metadata service for the user-data. By default the endpoint is reachable at "http://169.254.169.254/latest/user-data" without the slash "/" at the end.

For this reason see the ec2 metadata service documentation: http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html

It is also fixed in the coreOS repository, the source of this file:
https://github.com/coreos/coreos-cloudinit/blob/master/datasource/metadata/ec2/metadata.go

### DIFF
--- a/config/cloudinit/datasource/metadata/ec2/metadata.go
+++ b/config/cloudinit/datasource/metadata/ec2/metadata.go
@@ -30,7 +30,7 @@ import (
 const (
 	DefaultAddress = "http://169.254.169.254/"
 	apiVersion     = "latest/"
-	userdataPath   = apiVersion + "user-data/"
+	userdataPath   = apiVersion + "user-data"
 	metadataPath   = apiVersion + "meta-data/"
 )
 


niusmallnan:
There is no impact, both ways are no problem:
```
rancher@ip-172-31-26-84:~$ curl http://169.254.169.254/latest/user-data
test from zhiborancher@ip-172-31-26-84:~$

rancher@ip-172-31-26-84:~$ curl http://169.254.169.254/latest/user-data/
test from zhiborancher@ip-172-31-26-84:~$
```

r0b2g1t:
Hello,

for sure it works on AWS but I tested it on Open Telekom Cloud (OTC) which use the ec2 metadata service too. And RancherOS can't retrieve user-data from the URL http://169.254.169.254/latest/user-data/ because it is not the right endpoint and deviates from the reference as I had described.
Second the change to this endpoint was not in the original repository of CoreOS where the source code come from.

It is possible to include the change to promote the the spread of RancherOS to other platforms like OTC?

niusmallnan:
Ok, I can merge this PR.

> It is possible to include the change to promote the the spread of RancherOS to other platforms like OTC?

Sorry, there is not much time to focus on other cloud platforms.

PR #2207, opened 2018-01-05
Remove arm32 support

niusmallnan:


zimme:
I'm not against this, but what's the background for removing arm32 support?

niusmallnan:
Fewer and fewer Linux distributions support 32-bit OS, we also pay more attention to the future.

PR #2277, opened 2018-03-06
cmd/power: Another shutdown command fix

vancluever:
It looks like some arguments for shutdown/halt/poweroff have been moved
to a conditional block that works off of how the command was actually
called. However, this value is derived from argv 0, without any sort of
normalization to make sure it matches the relative commands used to
determine how arguments are handled.

This has particular implications when power management commands are
called via absolute commands, as for example in the case of
`open-vm-tools` which calls `/sbin/shutdown -h now` specifically when
shutting down a system.

This corrects the situation by passing argv 0 through `filepath.Base`
before operating on it.

### DIFF
--- a/cmd/power/shutdown.go
+++ b/cmd/power/shutdown.go
@@ -27,7 +27,7 @@ func Shutdown() {
 	log.InitLogger()
 	app := cli.NewApp()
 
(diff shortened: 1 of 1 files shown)


niusmallnan:
LGTM

Issue #2284, opened 2018-03-12
Cloud-config doesn't seem to work with  encoding: "gz+b64" or encoding: "gzip+base64"


josmo:
**RancherOS Version: (ros os version)**
ros version v1.0.1
also tried version v1.2.0 from os image rancher/os:v1.2.0

**Where are you running RancherOS? (docker-machine, AWS, GCE, baremetal, etc.)**
AWS

While trying to use user-data with gzip/base64 encoding for the content it seems to ignore gzip.
  tried encoding set to gz+b64 or gzip+base64 
```
  - path: /home/rancher/vpn_server.config
    permissions: "0644"
    owner: root
    encoding: "gzip+base64"
    content: $(content in b64/gzip)
```
expected:
creates /home/rancher/vpn_server.config file

Actual:
creates a weird directory I imagined because it's ignoring the gzip portion.

Jason-ZW:
@josmo Are you sure can create a weird directory？ It should create nothing when set encoding "gzip+base64" base the current code logic. Could you please check and let me know system-docker log with the command `system-docker logs -f console` and it shoud be an error log such like this `level=error msg="Error writing file" err="Unable to write file with encoding gzip+base64" path= `?

josmo:
@Jason-ZW looks like you're correct. With gzip+base64 it does give that error. I know there was a scenario where it created a weird directory, but looks like I can't repro now.   Seems like 1.4 will support the gzip+base64 encoding at the file level :)

kingsd041:
@josmo 
It has been fixed in rancherosv1.4.0-rc1 and I have tested it

josmo:
@kingsd041 
Thanks a bunch!

PR #2304, opened 2018-03-26
Add decode logic for writefiles

Jason-ZW:
Relate to #2284 

CloudInit WriteFiles failed when user define user-data with encoding like this 
```yaml
#cloud-config
write_files:
 - path: /home/rancher/vpn_server.config
   permissions: "0644"
   owner: root
   encoding: "gzip+base64"
   content: $(content in b64/gzip)
 ```
Reason:
  In CoreOS cloudinit, user-data are parsed by `ParseUserData` function which contain a `cc.Decode()` logic in it.
  Our repo have the same function with coreOS `ParseUserData` but is not called anywhere.So I extracted the main logical part of it and added it here to solve the `WriteFiles` problem, ensure that the impact was minimized.

https://github.com/coreos/coreos-cloudinit/blob/master/coreos-cloudinit.go#L232
https://github.com/coreos/coreos-cloudinit/blob/3c68e2e67d2ea077e2b79383c0ad313770c62ca2/initialize/user_data.go#L44

### DIFF
--- a/cmd/cloudinitexecute/cloudinitexecute.go
+++ b/cmd/cloudinitexecute/cloudinitexecute.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	rancherConfig "github.com/rancher/os/config"
+	"github.com/rancher/os/config/cloudinit/config"
 	"github.com/rancher/os/config/cloudinit/system"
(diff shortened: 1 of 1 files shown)


niusmallnan:
LGTM

PR #2336, opened 2018-04-26
Added Raspberry Pi 3B+ support

### LINKED ISSUE #2327: Ethernet support on Pi3 B+
**RancherOS Version: 1.3.0**

**Where are you running RancherOS? Raspberry Pi 3 Model B+**

There has been a change of hardware on the Pi 3 B+ which prevents the use of the new ethernet port.

When adding the new .dtb, the boot fails.

I'm trying to find a workaround that doesn't involve a custom recompile of RancherOS.
If someone is interested by this issue, some help would be greatly appreciated 😄

olljanat:
Updated to latest version of rpi64-kernel which contains support for Raspberry Pi 3B+

Fixes #2327

### DIFF
--- a/scripts/images/raspberry-pi-hypriot64/Dockerfile.dapper
+++ b/scripts/images/raspberry-pi-hypriot64/Dockerfile.dapper
@@ -13,7 +13,7 @@ RUN mkdir -p /source/assets
 COPY rootfs_arm64.tar.gz /source/assets/rootfs_arm64.tar.gz
 
 ENV URL=https://github.com/DieterReuter/rpi64-kernel/releases/download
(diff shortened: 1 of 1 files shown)


niusmallnan:
@olljanat  Hi, not sure does it also need to update rpi-bootloader?
You have tested it out, right?

olljanat:
Ted I tested by copying new version of kernel8.img and that new bcm2710-rpi-3-b-plus.dtb files to both partitions on sd card. After that I was able to boot Rancher OS 1.3.0 on my Pi3B+ and eth0 started working.

What I have not tested is build whole OS from sctrach but I assume that you have good automation for it.

Issue #2283, opened 2018-03-12
RancherOS on ESXi 6.5: Waiting for SSH


ivanfilippov:
**RancherOS Version: (ros os version)**
v1.2.0

**Where are you running RancherOS? (docker-machine, AWS, GCE, baremetal, etc.)**
VMWare ESXi 6.5

I'm trying to launch RancherOS on ESXi 6.5 via the Rancher Web UI but I'm getting stuck on "Waiting for SSH". I've tried to launch an instance via docker-machine manually and get the same problem.

I believe the issue is that docker-machine is unable to inject its SSH keys into RancherOS, but I'm not sure where to go from there.

Does anyone have any ideas?

docker-machine command:
```
docker-machine -D create \
	--engine-registry-mirror "http://???" \
	--driver vmwarevsphere \
	--vmwarevsphere-username ???\
	--vmwarevsphere-password ???\
	--vmwarevsphere-vcenter ??? \
	--vmwarevsphere-network ??? \
	--vmwarevsphere-datastore ??? \
	--vmwarevsphere-disk-size 10000 \
	--vmwarevsphere-cpu-count 2 \
	--vmwarevsphere-memory-size 2048 \
	--vmwarevsphere-boot2docker-url "http://releases.rancher.com/os/latest/rancheros.iso" \
	--vmwarevsphere-cloudinit "https://???" \
	rancheros-test
````

 docker-machine debug log:
```
Docker Machine Version:  0.14.0, build 89b8332
Found binary path at /usr/local/bin/docker-machine
Launching plugin server for driver vmwarevsphere
Plugin server listening at address 127.0.0.1:37875
() Calling .GetVersion
Using API Version  1
() Calling .SetConfigRaw
() Calling .GetMachineName
(flag-lookup) Calling .GetMachineName
(flag-lookup) Calling .DriverName
(flag-lookup) Calling .GetCreateFlags
Found binary path at /usr/local/bin/docker-machine
Launching plugin server for driver vmwarevsphere
Plugin server listening at address 127.0.0.1:39809
() Calling .GetVersion
Using API Version  1
() Calling .SetConfigRaw
() Calling .GetMachineName
(rancheros-test) Calling .GetMachineName
(rancheros-test) Calling .DriverName
(rancheros-test) Calling .GetCreateFlags
(rancheros-test) Calling .SetConfigFromFlags
Reading certificate data from /root/.docker/machine/certs/ca.pem
Decoding PEM data...
Parsing certificate...
[...]

PR #2348, opened 2018-05-03
Create boot2docker directory to solved waiting for ssh problem when using docker-machine vmware related  driver

Jason-ZW:
This PR will fix the ssh problem when using `docker-machine` with the VMware related driver such as "vmwarefusion, vmwarevsphere...".


Jason-ZW:
related to #2283

Issue #2337, opened 2018-04-28
config exported using ros config export is invalid


leodotcloud:
```
[rancher@rancher ~]$ sudo ros config export
EXTRA_CMDLINE: /init
rancher:
  environment:
    EXTRA_CMDLINE: /init
  network:
    interfaces:
      eth0:
        address: 1.1.1.2/29
        dhcp: false
        gateway: 1.1.1.1
  password: blahblahblah
  services_include:
    open-vm-tools: true
  state:
    dev: LABEL=RANCHER_STATE
    wait: true
ssh_authorized_keys: []
```

```
[rancher@rancher ~]$ sudo ros config export > /tmp/c.yaml
[rancher@rancher ~]$ sudo ros config validate -i /tmp/c.yaml
ERRO[0000] EXTRA_CMDLINE: Additional property EXTRA_CMDLINE is not allowed
ERRO[0000] password: Additional property password is not allowed
[rancher@rancher ~]$
```

```
[rancher@rancher ~]$ system-docker ps -a
Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/system-docker.sock: Get http://%2Fvar%2Frun%2Fsystem-docker.sock/v1.31/containers/json?all=1: dial unix /var/run/system-docker.sock: connect: permission denied
[rancher@rancher ~]$ sudo system-docker ps -a
CONTAINER ID        IMAGE                              COMMAND                  CREATED             STATUS                     PORTS               NAMES
db6668ac0262        rancher/os-base:v1.3.0             "/usr/bin/ros entr..."   8 minutes ago       Exited (0) 8 minutes ago                       preload-user-images
b8ac2f88a31c        rancher/os-docker:17.09.1          "ros user-docker"        8 minutes ago       Up 8 minutes                                   docker
902bef4bc94a        rancher/os-console:v1.3.0          "/usr/bin/ros entr..."   8 minutes ago       Up 8 minutes                                   console
4a75892d0c38        rancher/os-base:v1.3.0             "/usr/bin/ros entr..."   8 minutes ago       Exited (0) 8 minutes ago                       cloud-init-execute
fd3cc215bd00        rancher/os-base:v1.3.0             "/usr/bin/ros entr..."   8 minutes ago       Up 8 minutes                                   ntp
ec639f3b81d9        rancher/os-base:v1.3.0             "/usr/bin/ros entr..."   8 minutes ago       Up 8 minutes                                   network
[...]

PR #2351, opened 2018-05-07
Add export filter of additional property

wchao1241:
fix issue [#2337](https://github.com/rancher/os/issues/2337)

Add the property filter that filter out illegal fields


niusmallnan:
LGTM

PR #2379, opened 2018-06-05
Switch to Cloudflare default DNS

jlelse:
The Cloudflare DNS servers are probably better for more speed & privacy.


niusmallnan:
Does `docker/libnetwork` use the Cloudflare DNS as default?

jlelse:
Oh no it uses Google. So I will try to send a PR there .

PR #2393, opened 2018-06-22
Decompress user-data if it is gzipped

### LINKED ISSUE #2391: Cloud-Init which is gzipped won't work
1.4.0

Exoscale, CloudStack, all cloud-init providers which fetch userdata

Accordingly to [cloud-init documentation](http://cloudinit.readthedocs.io/en/latest/topics/format.html#gzip-compressed-content) the content can be gzip compressed. There are some providers which will compress the user-data especially because it could be limited to [~16384 bytes](http://cloudinit.readthedocs.io/en/latest/topics/format.html#gzip-compressed-content) with some providers.
In all these cases RancherOS won't be able to load and apply the cloud-config, because it makes no attempt to uncompress gzipped streams.

stffabi:
Fixes #2391

### DIFF
--- a/cmd/cloudinitsave/cloudinitsave.go
+++ b/cmd/cloudinitsave/cloudinitsave.go
@@ -182,6 +182,11 @@ func fetchAndSave(ds datasource.Datasource) error {
 		log.Errorf("Failed fetching user-data from datasource: %v", err)
 		return err
 	}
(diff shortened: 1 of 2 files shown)


Jason-ZW:
@stffabi  LGTM

Issue #2177, opened 2017-12-15
Enhance ROS to create multiple user-docker daemons


jambajaar:
We wish to fully launch Rancher and it's agents from Rancher stacks for ease of testing Rancher itself.

E.g. 
* Given Rancher Instance "A" with an Environment "A" with 3 hosts,
* Creating a stack on Environment A with rancher/server and the rancher/agent,
* 1 host in Environment "A" runs the Rancher instance "B", with Environment "B"
* 2 hosts in Environment "A" run rancher/agents that connect to Environment :B:

This allows us to run test automation against Rancher instance B without creating dedicated VMs.

We did a spike to investigate feasibility, but as expected found the rancher/agent for Environment "B" collides with the agent running for Environment "A". 

Our request to Rancher is to provide a solution for our use-case. Two options we saw (in order of preference)
* Enhance ROS to create multiple user-docker daemons
* Enhance the rancher/agent to allow multiple simultaneous instances

niusmallnan:
@jambajaar 
Under the scenario you are talking about, Env A and B will use the system iptable on the same host, which may bring a lot of problems.

niusmallnan:
The users want a `Rancher in Rancher` case.
![image](https://user-images.githubusercontent.com/998984/36184878-4b37fb10-1171-11e8-8073-96783c126261.png)

The user-docker-1 starts root Rancher, and user-docker-2 starts child Rancher.
The user-docker-1 and user-docker-2 are shared host network stacks, so also share the same iptables.

This usage has the following problems:
1. There will be multiple network-managers to handle iptables rules, there is a great chance of conflict.
2. All agents share the `/var/lib/rancher` directory, we know that this directory holds the agent information. There is also a great chance of conflict.

Even if RancherOS can support multiple user-docker, but to fully meet the needs of users also need to modify Rancher itself.

I think that creating dedicated VMs is actually simple enough, we should not make the problem more complicated.

niusmallnan:
Here are my steps about creating another user-docker daemon:
1.  Make a persistent storage directory(`/var/lib/docker2`) for this user-docker daemon.
```
#update /usr/share/ros/os-config.yml

user-volumes:
      ....
      ....
      volumes:
      - /home:/home
      - /opt:/opt
      - /var/lib/kubelet:/var/lib/kubelet
      - /var/lib/docker2:/var/lib/docker2
```

[...]

PR #2403, opened 2018-07-06
Add support for multi docker daemon

Jason-ZW:
Dind:
- [x] `ros engine create` can create dind engine
- [x] `ros engine remove` can remove dind engine
- [x] `ros service list` can list dind service
- [x] `ros service up` can up dind service
- [x]  user can run `docker-<dind name>` like use docker command
- [ ] user can ssh to dind container
- [ ] ~lb ports support for dind service~

### DIFF
--- a/cmd/control/engine.go
+++ b/cmd/control/engine.go
@@ -2,21 +2,38 @@ package control
 
 import (
 	"fmt"
+	"io/ioutil"
+	"net"
+	"os"
+	"os/user"
+	"path"
 	"sort"
+	"strconv"
 	"strings"
 
 	"golang.org/x/net/context"
 
+	yaml "github.com/cloudfoundry-incubator/candiedyaml"
 	"github.com/codegangsta/cli"
 	"github.com/docker/docker/reference"
+	"github.com/docker/engine-api/types"
+	composeConfig "github.com/docker/libcompose/config"
 	"github.com/docker/libcompose/project/options"
+	composeYaml "github.com/docker/libcompose/yaml"
+	"github.com/pkg/errors"
 	"github.com/rancher/os/cmd/control/service"
+	"github.com/rancher/os/cmd/control/service/app"
 	"github.com/rancher/os/compose"
 	"github.com/rancher/os/config"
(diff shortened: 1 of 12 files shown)


cmd/control/engine.go
```diff
@@ -36,6 +53,66 @@ func engineSubcommands() []cli.Command {
...
+					Usage: "set the fix ip for the engine",
+				},
+				cli.IntFlag{
+					Name:  "ssh-port",
+					Value: randomSSHPort(),
+					Usage: "set the ssh port for the engine",
+				},
+				cli.StringFlag{
+					Name:  "authorized_keys",
+					Value: authorizedKeysPath(),
```
niusmallnan: Recommended to write like this, better readability
```
authorizedKeys := c.String("authorized_keys")
if  authorizedKeys == ""{
    authorizedKeys = authorizedKeysPath()
}
```

cmd/control/engine.go
```diff
@@ -166,3 +329,206 @@ func CurrentEngine() (engine string) {
...
+		return errors.New("Must specify fix ip")
+	}
+
+	err = CheckUserDefineIPv4Address(fixIP, *userDefineNetwork)
+	if err != nil {
+		return err
+	}
+
+	isVersionMatch := false
+	for _, v := range MultiEngineVersions {
```
niusmallnan: If the versions do not match, we should expose an error.
Let uses know the exact reason.

cmd/control/engine.go
```diff
@@ -19,6 +31,11 @@ import (
 	"github.com/rancher/os/util/network"
 )
 
+var (
+	MultiEngineVersions = []string{"docker:17.12.1-dind", "docker:18.03-dind", "docker:18.03.1-dind"}
```
niusmallnan: For `docker:17.12.1-dind`, what is the meaning of `docker`?

niusmallnan: It's better to use `SupportedEngineVersions`

niusmallnan:
https://github.com/rancher/os/issues/2177

Issue #2408, opened 2018-07-12
Service is deleted when cancelling command or stating no.


adingilloRancher:
**RancherOS Version: (ros os version)**
v1.4

**Where are you running RancherOS? (docker-machine, AWS, GCE, baremetal, etc.)**
AWS

Command is still deleting service regardless of approval. 

```
[root@ip-172-31-200-169 ~]# ros service up dind1
INFO[0000] Project [os]: Starting project               
INFO[0000] [0/19] [dind1]: Starting                     
INFO[0000] Recreating dind1                             
INFO[0000] [1/19] [dind1]: Started                      
INFO[0000] Project [os]: Project started                
[root@ip-172-31-200-169 ~]# system-docker ps
CONTAINER ID        IMAGE                              COMMAND                  CREATED             STATUS              PORTS                             NAMES
51806a6c674e        cnrancher/docker:17.12.1-dind      "docker-entrypoint..."   5 seconds ago       Up 3 seconds        2375/tcp, 0.0.0.0:38413->22/tcp   dind1
81142a036b7a        rancher/os-docker:17.12.1          "ros user-docker"        23 minutes ago      Up 23 minutes                                         docker
d86137a0719a        rancher/os-console:7ef84fb         "/usr/bin/ros entr..."   25 minutes ago      Up 25 minutes                                         console
e3fe1e9a070d        rancher/os-base:7ef84fb            "/usr/bin/ros entr..."   25 minutes ago      Up 25 minutes                                         ntp
4fc595129bed        rancher/os-base:7ef84fb            "/usr/bin/ros entr..."   25 minutes ago      Up 25 minutes                                         network
0796089ddd01        rancher/os-base:7ef84fb            "/usr/bin/ros entr..."   25 minutes ago      Up 25 minutes                                         udev
ace3150c078f        rancher/os-acpid:7ef84fb           "/usr/bin/ros entr..."   25 minutes ago      Up 25 minutes                                         acpid
bb8f29638821        rancher/os-syslog:7ef84fb          "/usr/bin/entrypoi..."   25 minutes ago      Up 25 minutes                                         syslog
06b8792c508b        rancher/container-crontab:v0.4.0   "container-crontab"      35 minutes ago      Up 25 minutes                                         system-cron
[root@ip-172-31-200-169 ~]# ros engine rm dind1
Going to remove 51806a6c674ec7c58439f56bb5c9e0d0e916fe536f8dc98f3fce8e29522049b4
Are you sure? [yN]
^C
[...]

PR #2410, opened 2018-07-16
Fix bugs for multiple user docker daemons

niusmallnan:
1. Optimize the remove logic of the dind engine
2. Distinguish between user docker engine and Dind engine

https://github.com/rancher/os/issues/2408


niusmallnan:
https://github.com/rancher/os/issues/2177

Issue #2414, opened 2018-07-17
Kernel Too old


itstamil:
I'm trying to build the rancher os from source on a ubuntu machine as my base 
the steps followed are 
1) cloned the master version on [github](www.github.com/rancher/os) 
2) simply run make 
3) the output is like kernel too old

OMG its always showing me like that

niusmallnan:
You should upgrade your kernel, the version must be at least https://github.com/rancher/os/blob/master/Dockerfile.dapper#L67

itstamil:
thanks for reply
i have attend that change in my Dockerfile.dapper but no improvement in the error
it's always on Kernel too old

superseb:
The file shouldn't be changed, your kernel should be upgraded. Please share output of the following commands on the host you are running `make`:

- `uname -r`
- `docker info`

itstamil:
**uname -r** result shows like this

4.13.0-45-generic

**docker info** result show like this


Containers: 4
 Running: 1
 Paused: 0
 Stopped: 3
Images: 184
Server Version: 18.03.1-ce
Storage Driver: overlay2
 Backing Filesystem: extfs
 Supports d_type: true
 Native Overlay Diff: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: bridge host macvlan null overlay
 Log: awslogs fluentd gcplogs gelf journald json-file logentries splunk syslog
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: 773c489c9c1b21a6d78b5c538cd395416ec50f88
runc version: 4fc53a81fb7c994640722ac585fa9ca548971871
init version: 949e6fa
Security Options:
 apparmor
 seccomp
  Profile: default
Kernel Version: 4.13.0-45-generic
Operating System: Ubuntu 16.04.4 LTS
OSType: linux
Architecture: x86_64
CPUs: 4
Total Memory: 7.7GiB
Name: infyn
ID: IYGI:DDKI:W5SY:NMQI:M75I:BGC4:Q2NU:KBV5:NWUY:MY7V:33D5:32T4
Docker Root Dir: /var/lib/docker
Debug Mode (client): false
Debug Mode (server): false
Registry: https://index.docker.io/v1/
Labels:
Experimental: false
Insecure Registries:
 127.0.0.0/8
Live Restore Enabled: false

WARNING: No swap limit support

superseb:
[...]

PR #2433, opened 2018-08-06
Avoid kernel too old error when building ROS

niusmallnan:
https://github.com/rancher/os/issues/2037
https://github.com/rancher/os/issues/2414
https://github.com/rancher/os/issues/1992

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -5,4 +5,4 @@ pipeline:
     volumes:
     - /var/run/docker.sock:/var/run/docker.sock
     commands:
-    - dapper ci
+    - KERNEL_CHECK=0 dapper ci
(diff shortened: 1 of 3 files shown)


segator:
still happens 
Host Kernel: 4.14.32-rancher2
RancherOS Kernel: 4.14.32-rancher3
Your host kernel must be greater than RancherOS KERNEL_VERSION
FATA[0263] exit status 1
Makefile:11: recipe for target 'default' failed

niusmallnan:
Hi @segator ,
That is the result I expect.
This is to prompt the user to compile with the correct kernel version in advance.

liyimeng:
hi @niusmallnan could please elaborate why do you make such enforcement? can it be >= instead of >?

niusmallnan:
@liyimeng You can use `KERNEL_CHECK=0 make`.

liyimeng:
Thanks @niusmallnan!  I read the reason in another page

PR #2442, opened 2018-08-14
Running "cloud-init" before "b2d env"

Jason-ZW:
related to rancher/os#442

### DIFF
--- a/init/init.go
+++ b/init/init.go
@@ -53,6 +53,10 @@ func loadModules(cfg *config.CloudConfig) (*config.CloudConfig, error) {
 		mounted[strings.SplitN(reader.Text(), " ", 2)[0]] = true
 	}
 
+	if util.GetHypervisor() == "hyperv" {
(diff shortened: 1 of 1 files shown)


init/init.go
```diff
@@ -334,6 +340,68 @@ func RunInit() error {
 			}
 			return cfg, nil
 		}},
+		config.CfgFuncData{"b2d env", func(cfg *config.CloudConfig) (*config.CloudConfig, error) {
```
niusmallnan: This part looks complicated and difficult to maintain.
The `UdevSettle ` also runs in `cloud-init`. If we put `cloud-init` before `b2d env`, will it be simpler?

niusmallnan:
Please use this commit msg:
```
Running "cloud-init" before "b2d env"
1. Users can configure `autoformat` through cloud-config
2. Udevsettle is executed in cloud-init, and some platforms such as hyper-v need to scan devices earlier.
```

Issue #2443, opened 2018-08-14
Support initcmd on boot


niusmallnan:
Although there is already `bootcmd`, it runs in `ros-bootstrap` process.
The `ros-bootstrap` does not run every boot, So we need a feature that allows the user to run one command every boot.

I suggest calling it `initcmd`.

olljanat:
Referring to #2487 it looks that scripts from cloud-config are run on every boot starting from 1.4.1.

PR #2452, opened 2018-08-20
[WIP] Add cloud-init cmd and script

niusmallnan:
https://github.com/rancher/os/issues/2443

### DIFF
--- a/config/types.go
+++ b/config/types.go
@@ -105,6 +105,8 @@ type CloudConfig struct {
 	Rancher           RancherConfig         `yaml:"rancher,omitempty"`
 	Runcmd            []yaml.StringandSlice `yaml:"runcmd,omitempty"`
 	Bootcmd           []yaml.StringandSlice `yaml:"bootcmd,omitempty"`
+	CloudInitcmd      []yaml.StringandSlice `yaml:"cloud_init_cmd,omitempty"`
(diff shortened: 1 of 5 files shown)


config/types.go
```diff
@@ -105,6 +105,8 @@ type CloudConfig struct {
 	Rancher           RancherConfig         `yaml:"rancher,omitempty"`
 	Runcmd            []yaml.StringandSlice `yaml:"runcmd,omitempty"`
 	Bootcmd           []yaml.StringandSlice `yaml:"bootcmd,omitempty"`
+	CloudInitcmd      []yaml.StringandSlice `yaml:"cloud_init_cmd,omitempty"`
```
olleolleolle: To be as compatible and harmonious with Bootcmd, Runcmd as possible, perhaps the yaml should read `cloud_initcmd` instead of `cloud_init_cmd`?

(This requires a change to `config/schema.go` in this PR, as well.)

olljanat:
@niusmallnan what is status of this one? I just noticed that we would need it to get /var/lib/docker mounted from another disk as cloud-init mount looks to happening too late on process.

niusmallnan:
@olljanat If you want to get `/var/lib/docker` mounted, you can use `runcmd` which runs before user-docker.
This PR is currently in low priority.

PR #2482, opened 2018-09-18
Move around code for better clarity

ibuildthecloud:
There is no functional change. This change is to enhance the readability of
the code but more importantly allow certain parts of RancherOS init sequence
to be used elsewhere.  The major change here is really to break apart init.go
into a series of packages.


ibuildthecloud:
I was curious that if there is no real change then why did ~140 lines of code get added.  My theory there is that I created a bunch of new packages which caused more lines of imports to get added.

niusmallnan:
Thanks! 
LGTM.

PR #2490, opened 2018-09-26
Support vSphere network protocol profiles

kinolaev:
Part of rancher/rancher#15817

Added `guestinfo` params:

- `interface.<n>.ip.<m>.netmask`
- `dns.servers`
- `dns.domains`

because vSphere doesn't provide IP address in CIDR format and have one field for DNS servers and one for DNS search paths.

### DIFF
--- a/config/cloudinit/datasource/vmware/vmware.go
+++ b/config/cloudinit/datasource/vmware/vmware.go
@@ -77,10 +77,24 @@ func (v VMWare) FetchMetadata() (metadata datasource.Metadata, err error) {
 		}
 		metadata.NetworkConfig.DNS.Nameservers = append(metadata.NetworkConfig.DNS.Nameservers, val)
 	}
+	dnsServers, _ := v.read("dns.servers")
+	for _, val := range strings.Split(dnsServers, ",") {
+		if val == "" {
+			break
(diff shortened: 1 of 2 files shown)


niusmallnan:
Hi @kinolaev ,
We have supported these [guestinfo](https://rancher.com/docs/os/v1.x/en/installation/running-rancheros/cloud/vmware-esxi/).
Is there any special reason why these can't meet your needs?

kinolaev:
Here is official documentation for [vSphere vApp properties](https://code.vmware.com/apis/196/vsphere#/doc/vim.vApp.PropertyInfo.html) and for [Network protocol profile](https://docs.vmware.com/en/VMware-vSphere/6.5/com.vmware.vsphere.networking.doc/GUID-4BEEFDA9-E6FF-4D28-B0F3-B02864D8795B.html). You can see that vSphere haven't type expression for passing IP in CIDR format. Same for DNS properties.

niusmallnan:
It seems that this feature makes sense.

I have already recommended it to my colleague, looking forward to the feedback of other PRs.

cjellick:
@niusmallnan who are we waiting on for an LGTM?

PR #2496, opened 2018-09-30
Update rancheros.ipxe

bharrisau:
When booting from EFI the `initrd` kernel parameter is required to avoid a panic.

`Kernel panic - not syncing: VFS: Unable to mount root fs on unknown-block(0,0)`


niusmallnan:
We have not officially supported EFI, I need to do some testing to verify if there are other problems.

Issue #2473, opened 2018-09-09
ROS boot hang on kernels (maybe>=4.14.36)


niusmallnan:
**RancherOS Version: (ros os version)**
v1.4.1-rc2
kernel 4.14.67

**Where are you running RancherOS? (docker-machine, AWS, GCE, baremetal, etc.)**
docker-machine + VirtualBox

It only appears on the first boot, this should be related to ros-bootstrap.

dmesg output:
```
[   28.271974] random: fast init done
[   28.504239] input: ImExPS/2 Generic Explorer Mouse as /devices/platform/i8042/serio1/input/input4
[   28.642856] e1000 0000:00:03.0 eth0: (PCI:33MHz:32-bit) 08:00:27:c0:a2:3e
[   28.643546] e1000 0000:00:03.0 eth0: Intel(R) PRO/1000 Network Connection
[   29.043658] e1000 0000:00:08.0 eth1: (PCI:33MHz:32-bit) 08:00:27:99:89:e0
[   29.044367] e1000 0000:00:08.0 eth1: Intel(R) PRO/1000 Network Connection
[  134.643987] random: crng init done
[  134.644523] random: 4 urandom warning(s) missed due to ratelimiting
[  134.889131] EXT4-fs (sda): mounted filesystem with ordered data mode. Opts: (null)
[  134.968344] udevd[51]: starting version 3.2.5
[  134.971086] udevd[52]: starting eudev-3.2.5
[  136.375081] EXT4-fs (sda): mounted filesystem with ordered data mode. Opts: (null)
```

It seems that linuxkit also hit this problem.
https://github.com/linuxkit/linuxkit/issues/3032

rootwuj:
RancherOS Version 1.4.1-rc3  9/17
Verified fixed

niusmallnan:
When I create a ros instance using docker-machine+vbox, ros-bootstrap will run mkfs to format the disk on the first boot. The mkfs relies on the generation of random numbers. A commit in kernel 4.14.36 will result in the inability to use random numbers in early boot, and mkfs will get stuck for a long time. It is caused by the empty entropy source.
```
https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=v4.14.36&id=6e513bc20ca63f594632eca4e1968791240b8f18
```

By adding a entropy daemon to fill the entropy pool can solve this issue, so  I run `rngd` in ros-boostrap.


About why I didn’t run rngd always, there are two reasons:
[...]

PR #2535, opened 2018-10-25
User /dev/urandom as random number input for rngd

niusmallnan:
As /dev/hwrng may not exist in some environments.

Of course, this is a really bad idea, since I am simple filling the kernel entropy pool with entropy coming from the kernel itself!
But this does not need to consider the user's hw rngd drivers.


niusmallnan:
https://github.com/rancher/os/issues/2473

PR #2568, opened 2018-11-16
Change dhcpcd master node startup function

Jason-ZW:
This PR is used to solve the problem of setting both eth0 and wlan0 from DHCP networks. Such as:
```
rancher:
  network:
    interfaces:
      eth0:
        dhcp: true
      wlan0:
         wifi_network: network1
    wifi_networks:
      network1:
        ssid: xxxx
        psk: xxx
```

### DIFF
--- a/pkg/netconf/netconf_linux.go
+++ b/pkg/netconf/netconf_linux.go
@@ -24,10 +24,11 @@ const (
 )
 
 var (
-	defaultDhcpArgs = []string{"dhcpcd", "-MA4"}
-	exitDhcpArgs    = []string{"dhcpcd", "-x"}
-	exitWpaArgs     = []string{"wpa_cli", "terminate"}
-	dhcpReleaseCmd  = "dhcpcd --release"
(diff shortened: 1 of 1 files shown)


pkg/netconf/netconf_linux.go
```diff
@@ -178,7 +179,12 @@ func ApplyNetworkConfigs(netCfg *NetworkConfig, userSetHostname, userSetDNS bool
 
 	wg := sync.WaitGroup{}
 
-	//apply network config
+	// Don't start any interfaces other than those specified on the command line.
+	// This allows dhcpcd to be started in Master mode.
+	// Then wait for subsequent dhcpcd commands to start each interface as required.
+	runInactiveDhcp()
```
niusmallnan: Even if eth0 is in static mode and there are no wifi networks, do we still need to run it?

Jason-ZW: I think we need. This has no effect on other processes when eth0 is static mode and there are no wifi networks.
It's only run dhcpcd daemon in Master mode and wait for subsequent dhcpcd command.

Issue #84, opened 2015-03-12
Make RancherOS work on Azure


sheng-liang:
Perhaps we need to create a system container for running the Azure python agent?

imikushin:
I'm on it. Expect something very soon

ibuildthecloud:
@imikushin Since we are using cloud-init datasources from CoreOS, there is support for wagent, but we just haven't enabled it in the code.  You just need to add it here https://github.com/rancherio/os/blob/master/cmd/cloudinit/cloudinit.go#L265.  I'm not sure what that code does, but just thought it might be worthwhile to point out.

metagen:
Sorry, I'm not getting it-- is cloudinit a deploy tool that hooks up to various cloud providers and gets rancheros running?  Pls let me know if that's correct, thank you :)!

wlan0:
@metagen Its not a deploy tool. cloud-init is a distribution package, that does early initialization of an instance on various cloud. https://cloudinit.readthedocs.org/en/latest/

cloud-init is used to setup ssh-keys, set hostname, write files and execute scripts on a freshly created instance. The data can be provided to cloud-init by various data sources, such as config drive (http://docs.openstack.org/user-guide/content/config-drive.html) for openstack. In case of Azure,  waagent is the data source (http://azure.microsoft.com/en-us/documentation/articles/virtual-machines-linux-agent-user-guide/)

yellow1912:
Any update on this issue? Would really be interesting to try it out :)

deniseschannon:
With this support, we need to make sure we add in the ability to pass in `user-data`.

deniseschannon:
@yellow1912 We currently have a RancherOS v0.3.1 image out in the Azure Resource Management marketplace. We've documented instructions on how to launch a VM using their portal. 

http://docs.rancher.com/os/running-rancheros/cloud/azure/

Flowman:
I tried to create a RancherOS image from the marketplace. When I select location I get not available in any region.

Noticed that RancherOS on Azure is configured for BYOL.

"Bring your own license (BYOL). Publisher implements their own mechanisms for securing the licensing of their product. Customers must contact the publisher directly to purchase a license to use the software."

Question is where do I get a license to be able to run RancherOS in Azure?

deniseschannon:
Are you using the new portal or the old portal? They switched over to the Azure Resource Management portal, which is where it can be found. 

Even though it says BYOL, there is no license once you get the VM started.

[...]

PR #2614, opened 2018-12-24
Add azure support

Jason-ZW:
#84


images/01-base/etc/waagent.conf
```diff
@@ -0,0 +1,81 @@
+# Windows Azure Linux Agent Configuration
```
niusmallnan: Which system service needs to use this `waagent.conf`?
The waagent service has its own `waegent.conf` file.

os-config.tpl.yml
```diff
@@ -21,6 +21,10 @@ rancher:
   ssh:
     daemon: true
   hypervisor_service: true
+  {{if eq "true" .AZURE_SERVICE -}}
```
niusmallnan:
```  
  
  services_include:
    {{if eq "true" .AZURE_SERVICE -}}
    waagent: true
    {{end -}}
```

scripts/layout-initrd
```diff
@@ -58,6 +58,9 @@ for file in $(ls .make-*); do
         ".make-hyperv")
             cache_services "h/hyperv-vm-tools.yml"
             ;;
+        ".make-azure")
+            cache_services "w/waagent.yml"
```
niusmallnan: Also `cache_services "h/hyperv-vm-tools.yml"` ?

PR #2637, opened 2019-01-09
Add prompt message to ros os upgrade

Jason-ZW:
#399

### DIFF
--- a/cmd/control/os.go
+++ b/cmd/control/os.go
@@ -185,6 +185,10 @@ func osUpgrade(c *cli.Context) error {
 		log.Fatalf("ros install / upgrade only supported on 'amd64', not '%s'", runtime.GOARCH)
 	}
 
+	if isExist := checkGlobalCfg(); !isExist {
+		log.Fatalf("ros upgrade cannot be supported")
+	}
+
 	image := c.String("image")
 
 	if image == "" {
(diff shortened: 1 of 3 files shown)


cmd/control/os.go
```diff
@@ -324,3 +328,10 @@ func getUpgradeURL() (string, error) {
 	cfg := config.LoadConfig()
 	return cfg.Rancher.Upgrade.URL, nil
 }
+
+func isUpgradeAvailable() bool {
+	if _, err := os.Stat("/proc/1/root/boot/global.cfg"); err == nil || os.IsExist(err) {
```
niusmallnan: Can we make a common func about `/proc/1/root/boot/global.cfg`?
I can see this condition in other places.

cmd/control/os.go
```diff
@@ -185,6 +185,10 @@ func osUpgrade(c *cli.Context) error {
 		log.Fatalf("ros install / upgrade only supported on 'amd64', not '%s'", runtime.GOARCH)
 	}
 
+	if !isUpgradeAvailable() {
+		log.Fatalf("ros upgrade only supported on the instance which running from installed disk")
```
niusmallnan: `ros upgrade cannot be supported`?

Issue #2631, opened 2019-01-06
config validate should fail on configfiles in windows format


ChrisMir:
**RancherOS Version: (ros os version)**
1.4.2 & 1.5.0

**Where are you running RancherOS? (docker-machine, AWS, GCE, baremetal, etc.)**
baremetal & virtualbox

This has kept me busy for quiet some time. I composed a cloud-config.yml with notepad++ including basic networking and my ssh key. Moved the file to the installation device, validated (correctly) and installed ros with it, only to find out my ssh key was not accepted.

After some time I opened the configfile with vi, when I saw ^M at the end of each line, windows format!
So, changed the format to unix, redo all the steps and everything worked as expected.

I'd suggest the config validation function would at least give a warning when it encounters a windowformatted file

kingsd041:
In the RancherOS v1.5.1-rc1 version, we added verification of the windows format.
```
[root@rancher rancher]# ros config validate -i windows-format-test.yml
FATA[0000] file format shouldn't contain CRLF characters
```
@ChrisMir Thank you for your feedback

PR #2657, opened 2019-01-23
Add CR/LF check

Jason-ZW:
#2631


cmd/control/config.go
```diff
@@ -290,13 +291,48 @@ func validate(c *cli.Context) error {
...
+	}
+	input, err := os.Open(inputFile)
+	if err != nil {
+		return nil, err
 	}
+	defer input.Close()
 	return ioutil.ReadAll(input)
 }
+
+func hasCRLF(fileName string) bool {
```
niusmallnan: Hard to read, can we make it simple?

Not sure if this is fine:
```
content, err := ioutil.ReadFile(filename)
return bytes.Contains(content, []byte{13, 10})
```

refer to: https://www.programming-books.io/essential/go/1d3abcf6f17c4186bb9617fa14074e48-normalize-newlines

cmd/control/config.go
```diff
@@ -290,13 +291,20 @@ func validate(c *cli.Context) error {
...
-		var err error
-		input, err = os.Open(inputFile)
-		if err != nil {
-			return nil, err
-		}
-		defer input.Close()
+	if inputFile == "" {
+		return nil, errors.New("input parameter can not be empty")
+	}
+	input, err := os.Open(inputFile)
```
niusmallnan: Do we also need `os.Open`? I think you just only need `ioutil.ReadFile` once.

Jason-ZW: We have to do that, or we'll block IO

PR #2667, opened 2019-02-07
Embed subscriber service

Jason-ZW:
ROS upgrade/subscriber service

### DIFF
--- a/os-config.tpl.yml
+++ b/os-config.tpl.yml
@@ -381,6 +381,25 @@ rancher:
       volumes:
       - /sys:/host/sys
       - /var/lib/system-docker:/var/lib/system-docker:shared
+    subscriber:
+      image: {{.OS_REPO}}/os-base:{{.VERSION}}{{.SUFFIX}}
+      command: os-subscriber
+      environment:
+      - IN_DOCKER=true
+      labels:
+        io.rancher.os.scope: system
+        io.rancher.os.after: console
+        cron.schedule: "0 0 3 * * ?"
+      privileged: true
+      pid: host
+      ipc: host
+      net: host
+      uts: host
+      volumes_from:
+      - command-volumes
+      - system-volumes
+      volumes:
+      - /sys:/host/sys
   system_docker:
     exec: true
     storage_driver: overlay2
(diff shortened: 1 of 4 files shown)


os-config.tpl.yml
```diff
@@ -381,6 +381,25 @@ rancher:
...
+      privileged: true
+      pid: host
+      ipc: host
+      net: host
+      uts: host
+      volumes_from:
+      - command-volumes
+      - system-volumes
+      volumes:
+      - /sys:/sys
```
niusmallnan: better to use `/sys:/host/sys`.

images/01-base/Dockerfile
```diff
@@ -39,7 +39,8 @@ RUN rm /sbin/poweroff /sbin/reboot /sbin/halt && \
     ln -s /usr/share/dhcpcd/hooks/10-wpa_supplicant /lib/dhcpcd/dhcpcd-hooks/ && \
     rm -f /usr/share/bash-completion/completions/* && \
     sed -i 1,10d /etc/rsyslog.conf && \
-    echo "*.*                /var/log/syslog" >> /etc/rsyslog.conf
+    echo "*.*                /var/log/syslog" >> /etc/rsyslog.conf && \
+    chmod +x /usr/bin/subscriber
```
niusmallnan: better to use `os-subscriber`

os-config.tpl.yml
```diff
@@ -399,6 +418,7 @@ rancher:
   upgrade:
     url: {{.OS_RELEASES_YML}}/releases{{.SUFFIX}}.yml
     image: {{.OS_REPO}}/os
+    policy: "download"
```
niusmallnan: Can consider removing double quotes.

images/01-base/Dockerfile
niusmallnan: And we don't need `chmod` if we pre-set permission for this file.

os-config.tpl.yml
Jason-ZW: Already modified

images/01-base/Dockerfile
Jason-ZW: Already modified

os-config.tpl.yml
Jason-ZW: Already modified

Issue #2677, opened 2019-02-15
Support for service cache updates


niusmallnan:
When we get services from `os-services` via `ros service`, ROS can download the files to cache, they will be saved to `/var/lib/rancher/cache`.

These caches are now unable to be updated, We should support updating the cached files.

We can add an option to `ros service list`, suck as:
```
$ ros service list --update
```
Also, need to apply `engine` and `console` and `os` index.

rootwuj:
Tested with rancher/os:v1.5.2-rc1 from May 22

Run the following command can update the cache file (/var/lib/rancher/cache/), and other files are not affected.
*  sudo ros service list --update
*  sudo ros console list --update
*  sudo ros engine list --update
*  sudo ros os list --update

The following command can work with the updated cache file.

*  sudo ros service up
*  sudo ros console switch
*  sudo ros engine switch
*  sudo ros os upgrade

After the reboot or shutdown, everything went as expected.

Verified fixed

PR #2688, opened 2019-02-26
Support service cache update

Jason-ZW:
## Features(#2677)
Existing cache files in this directory`(/var/lib/rancher/cache/)` will be updated, other files will not be affected.
- [x] `ros service list --update`
- [x] `ros console list --update`
- [x] `ros engine list --update`
- [x] `ros os list --update`

## Test Points
- Whether the `ros service list --update` is effective.
- Whether the `ros console list --update` is effective.
- Whether the `ros engine list --update` is effective.
- Whether the `ros os list --update` is effective.
- Whether the `ros service up` can  work with the updated cache file.
- Whether the `ros console switch` can  work with the updated cache file.
- Whether the `ros engine switch` can  work with the updated cache file.
- Whether the `ros os upgrade` can  work with the updated cache file.
- After the reboot everything went as expected

### DIFF
--- a/cmd/control/console.go
+++ b/cmd/control/console.go
@@ -43,8 +43,14 @@ func consoleSubcommands() []cli.Command {
 			Action: consoleEnable,
 		},
 		{
-			Name:   "list",
-			Usage:  "list available consoles",
+			Name:  "list",
+			Usage: "list available consoles",
+			Flags: []cli.Flag{
+				cli.BoolFlag{
+					Name:  "update, u",
+					Usage: "update console cache",
+				},
+			},
 			Action: consoleList,
 		},
 	}
@@ -127,7 +133,7 @@ func consoleEnable(c *cli.Context) error {
 
 func consoleList(c *cli.Context) error {
 	cfg := config.LoadConfig()
-	consoles := availableConsoles(cfg)
+	consoles := availableConsoles(cfg, c.Bool("update"))
 	CurrentConsole := CurrentConsole()
 
 	for _, console := range consoles {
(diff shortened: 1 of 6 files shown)


cmd/control/console.go
```diff
@@ -127,7 +133,11 @@ func consoleEnable(c *cli.Context) error {
 
 func consoleList(c *cli.Context) error {
 	cfg := config.LoadConfig()
-	consoles := availableConsoles(cfg)
+	isUpdate := false
+	if c.Bool("update") {
+		isUpdate = true
+	}
+	consoles := availableConsoles(cfg, isUpdate)
```
niusmallnan: `consoles := availableConsoles(cfg, c.Bool("update"))` ?

cmd/control/engine.go
```diff
@@ -272,7 +278,11 @@ func engineEnable(c *cli.Context) error {
 
 func engineList(c *cli.Context) error {
 	cfg := config.LoadConfig()
-	engines := availableEngines(cfg)
+	isUpdate := false
+	if c.Bool("update") {
+		isUpdate = true
+	}
+	engines := availableEngines(cfg, isUpdate)
```
niusmallnan: Same here

cmd/control/os.go
```diff
@@ -131,7 +144,11 @@ func getImages() (*Images, error) {
 }
 
 func osMetaDataGet(c *cli.Context) error {
-	images, err := getImages()
+	isUpdate := false
+	if c.Bool("update") {
+		isUpdate = true
+	}
+	images, err := getImages(isUpdate)
```
niusmallnan: Same here

pkg/util/network/cache.go
```diff
@@ -43,3 +43,11 @@ func cacheAdd(location string, data []byte) {
 	cacheFile := cacheDirectory + locationHash(location)
 	os.Rename(tempFile.Name(), cacheFile)
 }
+
+func cacheRemove(location string) error {
+	cacheFile := cacheDirectory + locationHash(location)
+	if err := os.Remove(cacheFile); err != nil {
```
niusmallnan: `return os.Remove(cacheFile)` ?

cmd/control/console.go
Jason-ZW: Already Modified

cmd/control/os.go
Jason-ZW: Already Modified

pkg/util/network/cache.go
Jason-ZW: Already Modified

cmd/control/engine.go
Jason-ZW: Already Modified

Issue #2676, opened 2019-02-14
/var/run/utmp is a directory ?


jbrockopp:
**RancherOS Version: (ros os version)**

```
[rancher@server ~]$ sudo ros -v
version v1.4.0 from os image rancher/os:v1.4.0
```

**Where are you running RancherOS? (docker-machine, AWS, GCE, baremetal, etc.)**

OpenStack

I think I found a bug so I'm opening this to get confirmation of expected behavior in regards to `/var/run/utmp` is no longer a file but a directory if you switch your console from the default.

I have two servers, one with the `default` console enabled and one with the `centos` console enabled.

Default Console:

```
[rancher@server ~]$ sudo ros console list
disabled alpine
disabled centos
disabled debian
current  default
disabled fedora
disabled ubuntu

[rancher@server ~]$ sudo ros -v
version v1.4.0 from os image rancher/os:v1.4.0

[rancher@server ~]$ ls -al /var/run/utmp
-rw-r--r--    1 root     root           768 Feb 14 15:58 /var/run/utmp
```

Centos Console:

```
[rancher@server ~]$ sudo ros console list
disabled alpine
current  centos
disabled debian
disabled default
disabled fedora
disabled ubuntu

[rancher@server ~]$ sudo ros -v
version v1.4.0 from os image rancher/os:v1.4.0

[rancher@server ~]$ ls -al /var/run/utmp/
total 0
drwxr-xr-x  2 root root  40 Jan 20 00:48 .
drwxrwxrwt 21 root root 680 Jan 20 02:25 ..
```

Is it expected behavior that when you switch your console to something besides `default`, it converts the `/var/run/utmp` to a directory? Or is this a bug?

I'll post in the next comment how this is affecting me.

jbrockopp:
The issue is we're deploying [Telegraf](https://hub.docker.com/_/telegraf) as a docker container as a part of our cloud-config.

Here is the relevant configuration:

```yaml
    telegraf:
      container_name: telegraf
      image: telegraf:1.9.4-alpine
      restart: always
      volumes:
        - /home/rancher/telegraf:/etc/telegraf:ro
        - /sys:/rootfs/sys:ro
        - /proc:/rootfs/proc:ro
        - /var/run/utmp:/var/run/utmp:ro
        - /var/run/docker.sock:/var/run/docker.sock:ro
```

The issue this produces in the logs is:

```
[...]

PR #2701, opened 2019-03-06
Add /run/utmp file

Jason-ZW:
#2676

### DIFF
--- a/cmd/control/console_init.go
+++ b/cmd/control/console_init.go
@@ -70,6 +70,15 @@ func consoleInitFunc() error {
 	createHomeDir(rancherHome, 1100, 1100)
 	createHomeDir(dockerHome, 1101, 1101)
 
+	// who & w command need this file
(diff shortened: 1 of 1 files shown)


cmd/control/console_init.go
```diff
@@ -70,6 +70,15 @@ func consoleInitFunc() error {
 	createHomeDir(rancherHome, 1100, 1100)
 	createHomeDir(dockerHome, 1101, 1101)
 
+	// who & w command need this file
+	if _, err := os.Stat("/run/utmp"); os.IsNotExist(err) {
+		f, err := os.OpenFile("/run/utmp", os.O_RDWR|os.O_CREATE, 0644)
```
niusmallnan: Why not `os.Create`? Any special reason?

Jason-ZW: os.Create perm is 0666. I need 0644.

PR #2706, opened 2019-03-12
Added wildcard matching on MAC addresses.

mmriis:
It would be a nice feature to be able to wildcard match on mac adresses for interfaces. This is eg. useful for bonding where you would like all nics of a certain make or model to be added to the bond.

Using this feature you could do the following:
```
#cloud-config
rancher:
  network:
    interfaces:
      bond0:
        dhcp: true
        bond_opts:
          mode: 802.3ad
          lacp_rate: fast
          miimon: 100
        post_up:
          - ip route add 0.0.0.0/0 via 1.2.3.4
      # Add all Mellanox adapters to bond
      "mac=98:03:9b*":
        bond: bond0
```

Issue #2707, opened 2019-03-12
Support wildcard matching on MAC addresses


niusmallnan:
From @mmriis :

It would be a nice feature to be able to wildcard match on mac adresses for interfaces. This is eg. useful for bonding where you would like all nics of a certain make or model to be added to the bond.

Using this feature you could do the following:

```
#cloud-config
rancher:
  network:
    interfaces:
      bond0:
        dhcp: true
        bond_opts:
          mode: 802.3ad
          lacp_rate: fast
          miimon: 100
        post_up:
          - ip route add 0.0.0.0/0 via 1.2.3.4
      # Add all Mellanox adapters to bond
      "mac=98:03:9b*":
        bond: bond0
```

niusmallnan:
@kingsd041 Can you add automated tests for this feature? Please add to your to-do list. 😄

kingsd041:
@mmriis has added a go test for this, so there is no need to add it to the automated test.

rootwuj:
Tested with rancher/os:v1.5.2-rc1 from May 24 with VirtualBox

Test:
* Create VM and set up multiple NICs
* Execute `ros install -d /dev/sda -c cloud-config.yml `
```
$ vi cloud-config.yml
#cloud-config
rancher:
  network:
    interfaces:
      bond0:
        dhcp: true
        bond_opts:
          downdelay: "200"
          lacp_rate: "1"
          miimon: "100"
          mode: "2"
          updelay: "200"
          xmit_hash_policy: layer3+4
      mac=08:00:27*:
        bond: bond0
        dhcp: false
```
Result:
* All NICs are added to bond0.
* All NICs have the same mac address and are in the UP state.
* Can connect to VM via ssh.

Verified fixed.

PR #2735, opened 2019-04-08
Wildcard mac matching fix

mmriis:
#2706 has an issue where you have to manually run dhcpcd after the os is booted. That is because that PR didn't whether the interface was a bond or vlan interface like it should (that is the functionality when matching on strict MAC address).

I've corrected this and added testing for that specific case.


niusmallnan:
https://github.com/rancher/os/issues/2707

Issue #2747, opened 2019-04-28
Move the dappper build image from Xenial to Bionic


niusmallnan:
In order to follow-up maintenance, we need to migrate them ASAP.
This involves all of the RancherOS related repositories.

niusmallnan:
Multiprocessor (multithreading) support was added to xz in version 5.2.
Bionic uses xz 5.2.2, so this will make ROS compile faster.

niusmallnan:
Got problems about kexec and RPi build on arm64 after moving Bionic.
Keep this open.

PR #2756, opened 2019-05-05
Move docker base image from Xenial to Bionic

niusmallnan:
https://github.com/rancher/os/issues/2747


niusmallnan:
Got a problem of building RPi release, revert this PR.

PR #2760, opened 2019-05-07
Add azure cloud-init

Jason-ZW:


os-config.tpl.yml
```diff
@@ -84,6 +84,9 @@ rancher:
   cloud_init:
     datasources:
     - configdrive:/media/config-2
+    {{- if eq "true" .AZURE_SERVICE}}
```
niusmallnan: We don't need this by default, this can be added in os-packer.

niusmallnan:
https://github.com/rancher/os/issues/2762

Issue #2828, opened 2019-06-27
Do not delete cache files when ros service list --update filed.


Jason-ZW:
1.5.x

kingsd041:
Using `ros service list --update` in rancheros v1.5.3-rc1 does not delete the cache file.
If rancheros can access the internet, the latest service list will be updated. If it is air gap network, the cache file will be kept.

PR #2830, opened 2019-07-01
Fix update cache bug

Jason-ZW:
#2828


pkg/util/network/network.go
```diff
@@ -230,12 +230,24 @@ func UpdateCaches(urls []string, key string) error {
 }
 
 func UpdateCache(location string) ([]byte, error) {
-	if err := cacheRemove(location); err != nil {
+	// move cache file to temp directory
+	tempFile, err := cacheMove(location)
+	if err != nil {
 		return []byte{}, err
```
niusmallnan: Here we can `return nil, err` ?

PR #2891, opened 2019-09-11
Add Exoscale provider for Cloudinit

pierre-emmanuelJ:
# Add Exoscale provider for Cloudinit

We make this PR to fix  a problem about `datasource`.
We are based on a Cloudstack fork so we have difference for the Cloudinit, 
especially for the `metadata` service endpoint.

So we solve that by implementing that for our own provider.


niusmallnan:
https://github.com/rancher/os/issues/2897

PR #2895, opened 2019-09-17
Fix Exoscale Cloudinit datasource race condition

pierre-emmanuelJ:
This PR fix a race condition in the Cloudinit Exoscale provider.
The problem was at datasource level, when it tries to get the datasource, sometime the instance network interface is not up or starting. 
So the request was failing.


niusmallnan:
https://github.com/rancher/os/issues/2897

Issue #2972, opened 2020-03-10
Update linux kernel to fix cpu quota


curantes:
**RancherOS Version: (ros os version)**
v1.5.5
**Where are you running RancherOS? (docker-machine, AWS, GCE, baremetal, etc.)**
VMware iso

There is a bug in the version of linux kernel that is in latest Rancher OS which cause cpu limits not to work as intended which faulty cpu throttling.

4.14.154 or newer of the 4.14 linux tree has this patched.

some info about this issue in a blog: https://medium.com/omio-engineering/cpu-limits-and-aggressive-throttling-in-kubernetes-c5b20bd8a718

info about patch in linux mailinglist: https://lore.kernel.org/stable/1573236387-15456-2-git-send-email-chiluk+linux@indeed.com/

github issue from kubernetes: https://github.com/kubernetes/kubernetes/issues/67577

curantes:
@niusmallnan Can you please look into updating the kernel?

josmo:
It would be great to see the 5.6+ kernel with wireguard in the next kernel update release if that's at all in the plans.

PR #2991, opened 2020-04-22
Bump kernel 4.14.176-rancher

niusmallnan:
https://github.com/rancher/os/issues/2972


igorcezar:
Will this kernel be available in some RancherOS release? 
Or there will be no more releases?

PR #3054, opened 2021-03-30
Add maintain only as essential notice

### LINKED ISSUE #3000: Is this project still alive?
There has only been 5 commits since the last release and even that release was running a fair amount of older software. 
* Kernel: 4.14 (2017 LTS but sad that it is still being used #2750 #2728 #2972 )
* SSH: 7.9 (2018)
* Docker: 19.03.5 (this is reasonable updated but there are several point releases since then that aren't available despite every previous version being available)

There are several major issues that don't appear to being addressed.
* UEFI boot #2251 
* IPv6 #2593 
* RaspberryPi 4 support #2875

There have been lots of commits to k3os so it appears that is where Rancher is putting its efforts now. If that is the case it would be nice to know and know if there is a migration plan?

evilhamsterman:
This adds the the same notice as was added to the documentation in rancher/docs#3107 to the README for as much visibility to the status of RancherOS. This will close #3000

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@
 [![Docker Pulls](https://img.shields.io/docker/pulls/rancher/os.svg)](https://store.docker.com/community/images/rancher/os)
 [![Go Report Card](https://goreportcard.com/badge/github.com/rancher/os)](https://goreportcard.com/badge/github.com/rancher/os)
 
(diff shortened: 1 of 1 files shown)


olljanat:
@dweomer @niusmallnan any possibility to get this one merged?

cjellick:
finally added something to this effect in a different PR

PR #3076, opened 2022-10-11
Fix memory leak in selinux_linux.go

bduffany:
I have not tested this, but was searching for memory leaks using [this CS query](https://cs.github.com/rancher/os?q=%22%2C%20C.CString%22) and came across this instance in rancher/os. Since it's so widely used I figured it's worth at least pointing out.


bduffany:
Nevermind, I see this project is not actively maintained.

